PR #1, opened 2015-04-16
Bower dependency auto install

antiheld:


bower.json
```diff
@@ -26,10 +25,9 @@
   ],
   "license": "MIT",
   "private": false,
-  "devDependencies": {
+  "dependencies": {
```
jeremytrawls: we should split this so that jquery and select2 are still under devDependencies and bootstrap is under dependencies as we are currently bundling the src files for jquery and selec2 in with our techne.min.js dist file

config.json
```diff
@@ -1,5 +1,6 @@
 {
-    "bootstrap_path": "/bower_components/bootstrap",
+    "bower_path": "../../bower_components",
```
jeremytrawls: do we still  need to have the two separate config options here for bower and bootstrap?

README.md
```diff
@@ -28,4 +28,26 @@ Within the repo you'll find the following directories and files:
...
+
+and add in the file something like: 
+
+```
+// Use this file to point to the right folders
+// for correct paths for icon-fonts etc 
+
+@bootstrap-path: '../../bower_components/bootstrap'; // Path to bootstrap folder
+@bower-path: '../../bower_components';  // Path to bower_components folder
+@hytech-path: '../../bower_components/hyTechne'; // Path to hytech folder
```
jeremytrawls: this should be achievable by using the bower-path variable with our fixed hyTechne folder name as bower will always install it under the registered folder name, so I'm not sure that we really need to have the extra variable at this level.

gulpfile.js
```diff
@@ -70,27 +73,31 @@ gulp.task('iconfont',
                 .pipe(gulp.dest('src/less/components'));
             }
         )
-        .pipe( gulp.dest('dist/css/fonts/') )
```
jeremytrawls: I think we're still going to need to copy the fonts to the docs/kss/public/css/fonts folder for the live web version.  If we don't we'll have to have the server setup and configured to have the index of techne.yass.io pointing to the deep path and then navigating back in the folder structure could be problematice pending the server configuration

PR #3, opened 2015-04-24
Design/minor design updates

antiheld:


bower.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hyTechne",
-  "version": "0.0.2",
+  "version": "0.1.5",
```
jeremytrawls: let's update the version here to 0.1.6 to match the next release

gulpfile.js
```diff
@@ -296,7 +254,7 @@ gulp.task('watch',
 
 // iconfont, less, kss_bootrap_src
 
-gulp.task('dist', [ 'iconfont', 'less', 'styleguide', 'deploy']);
```
jeremytrawls: reason behind not wanting to run icon font when we run the dist task?

antiheld: The icon fontfiles are always getting recreated and comited even if the font isn't changing

PR #53, opened 2015-07-02
Feature/cce 110

SAPMontreal:
Code snippet expand/collapse

### DIFF
--- a/docs/kss/public/script.js
+++ b/docs/kss/public/script.js
@@ -45,7 +45,7 @@ $( document ).ready(function(){
 
     // off canvas menu
     $('[data-toggle="offcanvas"]').click(function () {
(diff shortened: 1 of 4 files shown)


docs/kss/public/script.js
```diff
@@ -45,7 +45,7 @@ $( document ).ready(function(){
 
     // off canvas menu
     $('[data-toggle="offcanvas"]').click(function () {
-        $('.row-offcanvas').toggleClass('active')
```
jeremytrawls: I would like for us to use the bootstrap toggle opposed to custom implementation here.

SAPMontreal: shall we merge?

PR #101, opened 2015-08-19
Add data card margin adjustment

antiheld:


src/less/components/new/patterns/_data-card.less
```diff
@@ -1,7 +1,61 @@
-// Example Card
+// DataCard(s)
 //
 // Markup:
-//<ul class="list-group">
```
jeremytrawls: Can you add an example of  vertically aligned cards  as well as an example of a card with the hover state.  I know that we have that in builder and we briefly spoke a while back about integrating it into Techne as well.

PR #121, opened 2015-08-31
Updated Global Header / Footer Elements + Global CSS Path

antiheld:


jeremytrawls:
@antiheld could you please merge with the latest of develop

PR #131, opened 2015-09-04
CCE-170: Removed the scrollable tabs section

liam-myrie:


src/less/components/_tabs-scrollable.less
```diff
@@ -29,7 +29,7 @@
 //     </div>
 // </div>
 //
-// Styleguide 5.2
```
jeremytrawls: you removed the documentation, but it looks like the styles are still here.  Is there a specific reason for this?

PR #134, opened 2015-09-10
Cce 86

liam-myrie:


jeremytrawls:
@liam-myrie  it looks good, but the product overlay needs to open on lick as well as mouse over

HybrisRawls:
@liam-myrie  please merge the latest of develop into the branch for deploy

PR #146, opened 2015-09-30
Feature/cce 78

taifunhaeberle:


docs/kss/index.html
```diff
@@ -17,7 +17,7 @@
   <link rel="stylesheet" href="public/y-techne/dist/css/techne.min.css">
 
 </head>
-<body class="kss-body has-left-nav kss-sidebar-nav-mobile-only">
```
HybrisRawls: are we intentionally removing this nav class?

PR #149, opened 2015-10-09
Bugfix/cce 243

taifunhaeberle:


docs/kss/public/y-techne/dist/js/select2.js
```diff
@@ -1,7 +1,7 @@
 /*
 Copyright 2012 Igor Vaynberg
 
-Version: 3.5.4 Timestamp: Sun Aug 30 13:30:32 EDT 2015
```
HybrisRawls: is there a sepcific reason we're going to an older version of select2 here?

PR #167, opened 2015-10-21
Feature/cce 254

saadmhybris:


jeremytrawls:
@saadmhybris  after reviewing the changes with Dieter we've decided that the only page that should have the grey background should be the commponents page. 

Once this change is made we can merge it in.  Thanks

saadmhybris:
@jeremytrawls - updated to only have background color on components page.

PR #188, opened 2015-10-28
added min-height to prevent the footer from climbing up.

saadmhybris:


saadmhybris:
I have done some digging and we have position fixed on the left nav which takes it out of the page flow completely. Tried using flexbox, but that is throwing of width off for other containers and causing lots of other issues.   

@taifunhaeberle - let me know if you have any ideas here.

taifunhaeberle:
It was fixed in a different branch by @HybrisRawls

PR #206, opened 2015-11-10
Anchor position update to fix header overlap issue.

saadmhybris:


docs/template/public/kss.less
```diff
@@ -1501,3 +1500,10 @@ body.hyComponentsPage{
 .error-message-table .wide-col {
     min-width: 260px;
 }
+
+.page{
```
taifunhaeberle: Please use more specific css classes for the styling of the style guide to avoid conflicts with the components.

saadmhybris: done.

PR #220, opened 2015-12-02
Feature/CCE 295 cleaned up

jensru:
moved infinite scroll to guidelines page


saadmhybris:
I ended up creating a class to wrap up my work. we can consider this done or Johannes can rename the class i created if he does not like it :)

PR #234, opened 2015-12-10
Added favicon into index.html template

jensru:
Added favicon into index.html template


docs/template/index.html
```diff
@@ -5,7 +5,23 @@
 <!--[if gt IE 8]><!--> <html class="no-js" lang="en"> <!--<![endif]-->
 <head>
   <meta charset="utf-8">
-
+  <link rel="apple-touch-icon" sizes="57x57" href="images/favicons/apple-touch-icon-57x57.png">
```
jeremytrawls: @jensru IE only supports absolute paths. Can you please update them?

jensru: ok, done

PR #235, opened 2015-12-10
CCE-334 Update text of package.json

jensru:


package.json
```diff
@@ -12,8 +12,9 @@
     "type": "git",
     "url": "https://github.com/SAP/techne.git"
   },
-  "author": "Hybris Software",
-  "license": "MIT",
+  "homepage": "http://techne.yaas.io/",
```
jeremytrawls: @jensru I'm speaking with @taifunhaeberle  and this doesn't seem to be correct.  Can you confirm wether this is the correct update that we should have?

jensru: What do you mean with not correct?

jensru:
Confirmed its Apache 2.0 like in the https://github.com/SAP/techne/blob/master/LICENSE

PR #238, opened 2015-12-10
CCE-325 : cookie bar

saadmhybris:
@antiheld 

I had to bump up the z-index from 99 to 103 so that it would work with our footer. we change our footer implementation and it's fixed at the bottom now.


antiheld:
@saadmhybris  That's fine

PR #247, opened 2015-12-28
moved error messages to devportal's Api Best Practices

oviboc-hybris:
moved error messages to devportal's Api Best Practices to which there already is a link. The extra "recommended message" column is also included.


oviboc-hybris:
closing it as there is a duplicate from me (experimenting.. ignore me :) )

PR #273, opened 2016-01-27
Feature/cce 267

saadmhybris:
Disabled card example.


src/less/components/_data-card.less
```diff
@@ -217,6 +221,15 @@ a.dataCard{
         color: @default-blue;
       }
     }
+
+    &.cardRowDisabled{
+        background-color: #d2dae4;
```
taifunhaeberle: Please use variables for colors other than white and add them to the vars file.

jeremytrawls: agreed

PR #274, opened 2016-01-27
Feature/cce 353

saadmhybris:
Update markup for card status


src/less/components/_data.less
```diff
@@ -63,7 +63,17 @@
 // Markup:
 //<p class="statusDataValue">
 //  <span class="active">
-//    Active
+//    Active Label
+//  </span>
+//</p> <br />
+//<p class="statusDataValue">
```
taifunhaeberle: Please check the task again. The task is to get rid off the statusDataValues and use the Bootstrap labels ( http://getbootstrap.com/components/#labels )

jeremytrawls: @jensru  can you please clarify the desired changes?

jensru: @jeremytrawls people where confused, that we used two versions of displaying color, inside cards. Do you have any concerns with removing "statusDataValue"?

saadmhybris: So I have replaced the statusDataValue with bootstrap label in the card examples and and also the Status Value Label. Are we all set here?

@jensru @taifunhaeberle

PR #275, opened 2016-01-27
Feature/cce 269

saadmhybris:
Remove bold from default label and added bold + * to required fields.


src/less/overwrites/_forms.less
```diff
@@ -49,13 +51,13 @@ textarea{
...
   background-color: #e1e6ec;
 }
 
-label.control-label.required::after{
-  content: '*';
+label.control-label.required{
   font-weight: bold;
-
+  &:after{
+    content: ' *';
```
jeremytrawls: @jensru  should there be a space after the label before the *?

jensru: I moved my answer to the related jira ticket CCE-269

saadmhybris: According to dieter, yes 

https://jira.hybris.com/browse/CCE-269?focusedCommentId=387325&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-387325

PR #276, opened 2016-01-27
Feature/cce 356

saadmhybris:
Enabled the textarea to be scalable.


src/less/overwrites/_forms.less
```diff
@@ -23,7 +23,7 @@ label{
 }
 
 textarea{
-  resize: none;
+  // resize: none;
```
taifunhaeberle: Please read the ticket accordingly: It said "Enable vertical resizing" not both ways. 
resize: vertical; is the proper way.

jeremytrawls: @saadmhybris as @taifunhaeberle pointed out you need to use a rule for vertical resize only.

Please try using: 

resize: vertical;

PR #277, opened 2016-01-28
Feature/cce 81

saadmhybris:
Added Tooltip and Popover Component example.


src/less/overwrites/_popover.less
```diff
@@ -13,51 +13,34 @@
...
-      }
-
-
-      .popover-title{
-        background: darken(@background, 5%);
-        color: @color;
-        border-bottom: @border;
-      }
-    }
+@y-popover-bg : #7f90a4;
```
taifunhaeberle: There is a variable @popover-bg; in our normal vars file already please use this instead of creating a css overwrite.

jeremytrawls:
@saadmhybris please see @taifunhaeberle's comments above

src/less/styles.less
```diff
@@ -186,16 +186,20 @@
...
 
 
-// Trees
-//
-// The tree component is used to display and manipulate hierarchical data, e.g. product categories.
+
+// Tooltip and Popover
 //
 // Styleguide 19
 @import "./components/_tree";
```
jeremytrawls: is the the right include for styleguide section 19?

src/less/styles.less
```diff
@@ -186,16 +186,20 @@
...
-// Trees
-//
-// The tree component is used to display and manipulate hierarchical data, e.g. product categories.
+
+// Tooltip and Popover
 //
 // Styleguide 19
-@import "./components/_tree";
-
+@import "./components/_tooltop-popover";
```
jeremytrawls: lets fix the name of the file :-)

_tooptip-popover

jeremytrawls:
@saadmhybris  the styling of the popover for tours is wrong and should share the same styling as the tooltip.  

Locally if you build the docs and look at http://localhost:8080/Patterns.html you'll see in the very first screenshot what that should look like.

src/less/overwrites/_popover.less
```diff
@@ -13,51 +13,19 @@
...
-        background: darken(@background, 5%);
-        color: @color;
-        border-bottom: @border;
-      }
-    }
+.popover {
+  color:#fff;
+  border: none;
+  border-radius: 0;
+  background-clip:border-box;
```
taifunhaeberle: We have no border, why do we need the border-box here?

src/less/overwrites/_popover.less
```diff
@@ -13,51 +13,19 @@
...
+  font-size: 15px;
```
taifunhaeberle: Please add a variable in the base variable set for this change as by default the popover font-size depends on @font-size-base.

src/less/overwrites/_popover.less
```diff
@@ -13,51 +13,19 @@
...
+  opacity: 1;
```
taifunhaeberle: Is there a reason to set the default opacity here?

src/less/overwrites/_popover.less
```diff
@@ -13,51 +13,19 @@
...
+  .popover-title{
+    border-bottom: none;
+    font-size: 19px;
```
taifunhaeberle: Same as for the default font-size, also maybe clarify if we can keep it depended on the used headline. ( h1 / h2 / h3 etc)

src/less/overwrites/_tooltip.less
```diff
@@ -3,6 +3,6 @@
   box-shadow:  0px 2px 4px 0px rgba(73, 101, 130, 0.25);
   max-width: 300px; 
   padding: 10px; 
-  font-size: 11px; 
+  font-size: 15px;
```
taifunhaeberle: @font-size-small set by default, please introduce a new variable in the variables file in the tooltip section

src/less/overwrites/_popover.less
```diff
@@ -13,51 +13,19 @@
...
+    padding:25px 25px 15/2px 25px;
```
taifunhaeberle: Any specific reason why we use subpixels here?

src/less/styles.less
taifunhaeberle: Right, it's not building right now ;)

src/less/overwrites/_popover.less
saadmhybris: none. rounded up to next full pixel which is 8px.

src/less/overwrites/_popover.less
saadmhybris: opacity reset is needed on .tooltip.in 

removed from here as it defaults to 1 already.

src/less/overwrites/_popover.less
saadmhybris: tested without it and didn't make a difference so, removed it.

src/less/overwrites/_tooltip.less
saadmhybris: ended up using @font-size-base as dieter was ok with going 1px down.

src/less/overwrites/_popover.less
saadmhybris: using @font-size-base for body and @font-size-large for header. removed hard coded px values.

saadmhybris:
@taifunhaeberle  - I have addressed all your questions on code. 

@jeremytrawls  - As Johannes said on the call, we'll have to address having shared style for tour bubble and popover in a separate ticket.

PR #297, opened 2016-03-29
add a new label-warning using default orange color

bvanalderweireldt:
labels have too few colors by default, only 3 available states and intermediate one would be an improvement.


HybrisRawls:
@bvanalderweireldt can you open this pull request against the develop branch? Thanks

PR #319, opened 2016-04-27
CCE-427 : Update help styling to be italic + updated help screen shot…

saadmhybris:
…s on patterns page.


src/less/components/_forms-validationandhelp.less
```diff
@@ -46,9 +46,9 @@
 }
 
 .help-icon-inline:before{
-  margin: 0;
+  margin: 5px 0 0 0;
   font-size: 25px; 
-  color:#7e858d;
+  color:#686f88;
```
HybrisRawls: this color needs to move to a variable

src/less/components/_forms-help.less
```diff
@@ -36,18 +36,18 @@
...
    margin-top: 5px;
    margin-bottom: 10px;
-
    cursor: default;
+   font-style: italic;
+
    .hyicon{
       font-size: 25px;
       line-height: 0;
       color: currentColor;
```
HybrisRawls: it doesn't look like the color variable is declared correctly

taifunhaeberle: That's fine, see http://caniuse.com/#feat=currentcolor

HybrisRawls:
in this case do we even need to define the color since it would inherit it?

src/less/components/_forms-validationandhelp.less
saadmhybris: done.

PR #321, opened 2016-04-28
Feature/cce 372

saadmhybris:


src/less/components/_tooltip-popover.less
```diff
@@ -27,4 +27,4 @@
 //   </div> 
 // </div>
 //
-// Styleguide 19.2
\ No newline at end of file
+// Styleguide 20.2
```
HybrisRawls: it doesn't look like you updates all of the style guide ID's in this file

saadmhybris: done.

PR #329, opened 2016-06-07
feature/CCE-366: updating modal styles

HybrisRawls:


HybrisRawls:
@taifunhaeberle  please review this when you get a chance.  Pretty minimal changes, so it shouldn't take long at all

PR #337, opened 2016-06-13
CCE-415 : component updates.

saadmhybris:


src/less/overwrites/_badges.less
```diff
@@ -15,6 +15,7 @@
 	padding: 5px 12px;
 	border-radius: 15px;
   border: 1px solid @default-greyblue-500;
```
HybrisRawls: fix indentations here

src/less/overwrites/_tables.less
```diff
@@ -220,7 +222,7 @@ table.techne-table {
             background: transparent!important;
             box-shadow: none;
             > td{
-                padding: 0px 0px 10px!important;
+                padding: 0px!important;
```
HybrisRawls: you need a space between the px and !important for IE browsers.

src/less/components/_data-card.less
```diff
@@ -117,12 +117,12 @@ a.dataCard{
   cursor: default;
   overflow: visible!important;
```
HybrisRawls: we should fix this one to

PR #355, opened 2016-06-28
CCE-445 : Update home page title in blue bar to SAP Hybris Techne

saadmhybris:


docs/markdown/index.md
```diff
@@ -4,7 +4,7 @@
             <div class="page-header hyHomeHeader text-center">
                 <div class="col-md-8 col-md-offset-2">
                     <img src="images/home/techne-logo.png" alt="techne" class="techne-logo">
-                    <h1>hybris techné*</h1>
+                    <h1>SAP Hybris Techné</h1>
                     <p>techne is the design guideline and component library for modern, mobile-first, user-centric experience design on YaaS that gives UX professionals the freedom and creativity to implement experiences from design to deployment FAST</p>
```
HybrisRawls: its' not mentioned in the ticket, but I think we should update the spelling of Techne in this paragraph and the description below to use the spelling 'techné' as it's used in the title and in the body text as so. 

What do you think @kwczopek ?

PR #369, opened 2016-07-21
Feature/cce 453

HybrisRawls:


src/less/overwrites/__overwrites.less
```diff
@@ -36,6 +36,15 @@ h6{
    }
 }
 
+h1{
+    font-weight: 400 !important;
```
taifunhaeberle: Please take a look at https://github.com/twbs/bootstrap/blob/master/less/type.less we should also add the .h1 there. Also i doubt we need the important? ;)

src/less/overwrites/__overwrites.less
```diff
@@ -36,6 +36,15 @@ h6{
...
+}
+
+
+h2{
+    font-weight: 600 !important;
```
taifunhaeberle: Please take a look at https://github.com/twbs/bootstrap/blob/master/less/type.less we should also add the .h2 there. Also i doubt we need the important? ;)

src/less/overwrites/_modal.less
```diff
@@ -8,7 +8,7 @@
 //     <div class="modal-content">
 //       <div class="modal-header">
 //         <button type="button" class="close" data-dismiss="modal">&times;</button>
-//         <h4 class="modal-title">Modal Header</h4>
+//         <h1 class="modal-title">Modal Header</h1>
```
taifunhaeberle: I'm not sure if this is a good idea, as the modal is part of the normal dom tree of a page. This means we end up with two <h1>'s even if the modal is not visible. Maybe here a special style for the modal header would be more fitting?

docs/template/public/less/_variables.less
```diff
@@ -22,4 +22,11 @@
 @dark-blue: #27394f;
 @side-nav-link: #dbe3eb;
 @top-nav-hover: #0376C6;
-@btn-type-color: #0b0080;
\ No newline at end of file
+@btn-type-color: #0b0080;
+@kss-green: #04ce26;
```
taifunhaeberle: We really should separate kss (techne page) variables form the actuall framework

docs/template/public/kss.less
```diff
@@ -1,4 +1,4 @@
-@import url(https://fonts.googleapis.com/css?family=Dosis:400,300,600,700);
+@import url(https://fonts.googleapis.com/css?family=Dosis:400,500,600);
```
taifunhaeberle: I doubt we need to include that at all here, as we have the techne from distribution folder included?

docs/template/public/less/_variables.less
HybrisRawls: good catch, this should be in the css variables

src/less/overwrites/__overwrites.less
HybrisRawls: done.

src/less/overwrites/__overwrites.less
HybrisRawls: done

src/less/overwrites/_modal.less
HybrisRawls: should we use a h2? The h4 doesn't really make sense to me

docs/template/public/kss.less
HybrisRawls: agreed

docs/template/public/less/_variables.less
taifunhaeberle: @saadmhybris Please move variables that are used exclusively for the page styling out of the normal variables file. into the kss.less. If you want to use pre-existing variables of the framework in the kss.less please make a import of the less file.

saadmhybris: @HybrisRawls @taifunhaeberle  - isn't this file only for kss site? I talked to jeremy about this and he said not to import frame work variables here to keep things simple.

taifunhaeberle: @saadmhybris that's right. My bad !

HybrisRawls:
@taifunhaeberle  is this one ready to merge?

src/less/overwrites/_type.less
```diff
@@ -1,5 +1,22 @@
+
+h1{
```
taifunhaeberle: You are lacking the .h1 classes from bootstrap ( and all following). Please have a look at the typefile of bootstrap

saadmhybris: done. @HybrisRawls this is ready for merge now i believe.

PR #395, opened 2016-08-22
CCE-453 : Left nav scroll spy

saadmhybris:


docs/template/public/main.js
```diff
@@ -5,17 +5,17 @@ jQuery(document).ready(function() {
     $menuItem = $menu.find('.kss-menu-item'),
     ref = $menu.data('kss-ref');
 
-    $menuItem.eq(ref).addClass('kss-active');
-     $('[data-toggle="offcanvas"]').click(function () {
-          $('.row-offcanvas').toggleClass('active')
-      });
+    $menuItem.eq(ref).addClass('active');
+    // $('[data-toggle="offcanvas"]').click(function () {
```
HybrisRawls: avoid checking in commented code.  Lets just remove it completely

docs/template/public/less/_side-nav.less
```diff
@@ -42,10 +42,11 @@
        
         &:hover {
             color: @blue_gd;
+            background-color: transparent;
         }
     }
 
-    .kss-active a {
```
HybrisRawls: @saadmhybris  is there a need to change it from css-active to active? This change is going to call for a full regression check of every page

saadmhybris: @HybrisRawls the class names was changed because that's the class bootstrap's scroll spy component adds and removes to keep tracking of the active node when scrolling.

docs/template/public/main.js
saadmhybris: done.

PR #411, opened 2016-09-21
CCE-528: Fix the underline on hover in IE and for pagination

nicandra:
Hi @jeremytrawls,

Here are the changes I made:
* no underline on hover and on focus for btn-link (appears mostly in IE)
* no underline on hover for pagination (there is already a change of color)

Thank you,
Andreea


nicandra:
i was in master; so please ignore; will do it again from develop

PR #445, opened 2016-10-13
Added copy text according to the tabs names

joseegm:


SAPMontreal:
this looks ok for me - as it is only documentation update

PR #447, opened 2016-10-13
Copy text change

joseegm:


saadmhybris:
this looks good. we can merge after @SAPMontreal id done reviewing.

SAPMontreal:
this looks like only doc update, right?
If so then go ahead

PR #454, opened 2016-10-20
Feature/cce 323 product thumbnail

saadmhybris:


saadmhybris:
This PR includes work for CCE-323 & CCE-417

PR #462, opened 2016-11-02
CCE-541 add SmartEdit icons

SAPMontreal:


joseegm:
Some small issues:

1. New icons don't convert when running "gulp iconfont"... I checked the svg code but I dunno what's the problem. Maybe is the size, they are 32x32, other ones are 133.2x133.2..
2. We have already an "add" icon, let's use the old one

joseegm:
Changed the icon filenames to be only one word.
SmartEdit icons are hidden on the page, since the style is different.

saadmhybris:
looks good.

PR #474, opened 2016-11-08
CCE-441: Narrow rows and bold clickable anchors

joseegm:


src/less/globals/themes/_techne.less
```diff
@@ -295,7 +295,7 @@
 //## Customizes the `.table` component with basic values, each used across all table variations.
 
 //** Padding for `<th>`s and `<td>`s.
-@table-cell-padding:            20px;
+@table-cell-padding:            10px;
```
saadmhybris: maybe we should change this to 10px 20px. this way we can keep the size padding at 20px? 10px feels a bit too close to the edge.

joseegm: Yep. Agreed.

PR #475, opened 2016-11-08
CC-438: remove ul, ol padding left on dataCards inside KSS content

joseegm:


docs/template/public/less/_kss-content.less
```diff
@@ -163,4 +163,12 @@
...
         }
     }
+
+    ul, ol 
+    {
+        &.dataCards 
+        {
+            padding: 0;
+        }
+    }
```
saadmhybris: We need to fix this on the cards component itself and not the kss site as it is an issue with the component that need to be fixed where ever is it used. ;) 

pls apply the padding-left: 0; in _data-card.less on .dataCards class

joseegm:
After reviewing the card usage on builder there will be no conflicts

PR #476, opened 2016-11-08
CCE-544: All starter pages now with the toolbar without margins

joseegm:


saadmhybris:
I added some spacing below the toolbar as it looked pretty tight after going full width with the toolbar. looks good.

PR #478, opened 2016-11-14
CCE-500: Changed the "Add Button" outside the table, to be consistent…

joseegm:
… with out pattern on the components page


joseegm:
Since this is a small change I will merge it. Also need the updated table to continue with the next issue CCE- 548

PR #479, opened 2016-11-14
Feature/cce 548 - Minor changes to the table pattern

joseegm:


saadmhybris:
@joseegm  - just pulled the latest and the checkboxes are still not working. also the long code lines are still there.

PR #482, opened 2016-11-15
CCE-529: Fixed nasty bootstrap bug on mobile

joseegm:


saadmhybris:
Tested it on iphone and it works@

PR #491, opened 2016-11-24
Feature/cce 405 423

joseegm:


joseegm:
Bunch of copy corrections

PR #501, opened 2016-12-02
Feature/cce 564: Release notes v1.5.2

joseegm:


joseegm:
Dunno if we can use Markdown for the release notes... If not we can change it back

PR #510, opened 2016-12-12
Feature/feedback menu

jakub-gawlas:
Added two new icons, `feedback` and `gethelp`, which are used in new component `feedback-menu`. To styling these icons in feedback-menu I created new component `button-icon` (but not sure if it should be component).
I will be very grateful for review ;)
Thanks!


src/less/components/_feedback-menu.less
```diff
@@ -0,0 +1,26 @@
...
+// <div class="feedback-menu">
+//  <span class="hyicon hyicon-feedback button-icon"></span>
+//  <span class="hyicon hyicon-gethelp button-icon"></span>
+// </div>
+//
+// Component 5.5
+
+.feedback-menu {
+  float: left;
+  font-size: 35px;
```
jakub-gawlas: menu should be stick to right-bottom edge of screen, so here should i add code responsible for that? 
```
position: relative;
z-index: 1000;
right: 0;
bottom: 100px;
```
or these values should be defined by user in own css?

saadmhybris:
@jakub-gawlas can you please rename the branch to Jira ticket number in techne backlog that this pull request corresponds to? it will help us in tracking the changes in case something goes wrong.

PR #516, opened 2016-12-16
CCE-558: Added the static use case

joseegm:


saadmhybris:
@joseegm - please see my comment on the ticket 

https://jira.hybris.com/browse/CCE-558

PR #536, opened 2017-01-04
Feature/cce 588 detail1 starter page update

like-me1:


src/less/techne-patterns/_avatars.less
```diff
@@ -212,7 +200,7 @@
 // <div class="row">
 //    <div class="col-xs-12 col-sm-5 col-md-4">
 //      <div class="thumbnail-shape center-block">
-//        <div class="thumbnail__img thumbnail-shape--circle thumbnail--avatar"></div>
+//        <div class="thumbnail__img thumbnail-shape--circle" style="background-image: url('/images/cat-avatar.jpg')"></div>
```
saadmhybris: why are we using inline style here?

src/less/techne-patterns/_avatars.less
```diff
@@ -323,7 +311,7 @@
 // <div class="row">
 //    <div class="ccol-xs-12 col-sm-5 col-md-4">
 //      <div class="thumbnail-shape center-block">
-//        <div class="thumbnail__img thumbnail-shape--rectangle thumbnail--org-logo"></div>
+//        <div class="thumbnail__img thumbnail-shape--rectangle" style="background-image: url('/images/sap-hybris-logo.png')"></div>
```
saadmhybris: why use inline style here?

src/less/techne-patterns/_avatars.less
```diff
@@ -431,7 +419,7 @@
 // <div class="row">
 //    <div class="col-xs-12 col-sm-5 col-md-4">
 //      <div class="thumbnail-shape center-block">
-//        <div class="thumbnail__img thumbnail-shape--rectangle thumbnail--package-logo "></div>
+//        <div class="thumbnail__img thumbnail-shape--rectangle" style="background-image: url('/images/package-picture.png')"></div>
```
saadmhybris: same thing here with inline style.

saadmhybris:
@like-me1  - when i click on the upload button, nothing happens. when i click on the delete button, the image cat image gets replaced with a spinner and the upload image dialog opens. Not sure if this is the right interaction?

src/less/techne-patterns/_avatars.less
like-me1: @saadmhybris this is the part of task description: * Change the sample background images from CSS classes to inline style: background-image. Because in YaaS usually the background image coming from media service, therefore we can't have it within the LESS files and we could wait quite long until it will be delivered to our app. So in AngularJS this syntax is used:ng-style="{'background-image':'url('+project.imageUrl+')'}" . So we wanted to be as much precise as possible.

like-me1:
@saadmhybris I fixed the issue, please don't forget to run gulp less command before running the server. The loading process and hidden buttons on click were not described in the task. It's just a logical outcome from my side.

PR #542, opened 2017-01-16
CCE-610: Changes suggested by FlySpeck added

joseegm:


saadmhybris:
learn some new grammar today.

PR #545, opened 2017-01-18
CCE-619 - Adding a cachebuster mechanism for the iconfont

taifunhaeberle:


joseegm:
@taifunhaeberle Don't be surprised you still assigned we are using the "Review" process as intended in github, so authors stay assigned to the pull requests.

PR #567, opened 2017-02-13
CCE-543, refactor the section inline level help tooltip classes

like-me1:
to reduce amount of classes used for styling the refactoring was made, mainly only for the section 5.8 (Section Inline Level help for read-only fields)

### DIFF
--- a/src/less/components/_forms-help.less
+++ b/src/less/components/_forms-help.less
@@ -200,7 +200,7 @@
         display: block;
         margin-bottom: 0;
     }
-    .form-readonly-text{
+    .form-readonly-text, .form-readonly-text__inline {
         display: inline-block;
     }
     .help-icon-inline {
--- a/src/less/components/_forms-help.less
+++ b/src/less/components/_forms-help.less
@@ -162,7 +162,10 @@
 
 .help-inline--section{
    &.help-inline--tooltip{
-      margin-bottom: 0;
+     margin-bottom: 0;
+     display: inline-block;
+     float: none;
+     position: relative;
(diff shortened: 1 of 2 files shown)


src/less/components/_forms-help.less
```diff
@@ -183,32 +186,26 @@
...
-// Component 5.7
+// Component 5.8
 
 .form-group__inline{
     label{
         display: block;
         margin-bottom: 0;
     }
-    .form-readonly-text__inline{
+    .form-readonly-text{
```
like-me1: @xak @joseegm  Good point, I just double checked with @saadmhybris and Core Builder Project and Alligators (Package and Operator view in Builder) and no-one use it and never used. Therefore I'll suggest to deprecate the .form-readonly-text__inline and announce it in release notes. Please let me know what you think?

SAPMontreal: what about the EC projects?

xak: Would it be easy enough to just leave the old class in place in the CSS even if the example markup and documentation don't mention it? 

```
.form-readonly-text, .form-readonly-text__inline { }
```

But I'll defer to you guys since you know better who is using what and what the impacts would be. It seems very low-risk to me at the end of the day.

joseegm: Agree on leaving the old classes in place for a couple versions, like we are doing with the localized-editor.

joseegm:
@xak I'm merging this, since it looks to be fine and it's seating here since a couple days now

PR #568, opened 2017-02-13
CCE-624

xak:
I was unable to update the following at this time: `gulp-less`, `select2`, `jquery`.

Most of the dependency updates were trivial, some were removed since they were no longer used. The most impactful was the `gulp-iconfont` module since major API updates have occurred. The task was rewritten and a `WOFF2` format was added to the output.


saadmhybris:
Looks good. Font's generate just fine.

joseegm:
Missed this message. We agreed to leave select2 out.
Testing this locally now

PR #570, opened 2017-02-20
CCE-662: Changed to display original and discounted prices vertically…

joseegm:
…, also to display price ranges

### DIFF
--- a/src/less/techne-patterns/_product-thumbnail.less
+++ b/src/less/techne-patterns/_product-thumbnail.less
@@ -131,6 +131,10 @@ Pattern 15.1
 
         .product-price-discounted {
             .text-danger;
+
+            .product-price-currency {
+                color: @techne-typo-default;
+            }
         }
     }
 
--- a/src/less/techne-patterns/_product-thumbnail.less
+++ b/src/less/techne-patterns/_product-thumbnail.less
@@ -28,9 +28,9 @@ Pattern 15.1
 */
 
 /*
-Display discounted pices
+Display discounted prices
 
-The Product Tumbnail can also display disconted prices. The original price will be displayed besides the discounted price. Both price have different styles to be easily recognizable.
+The Product Thumbnail can also display discounted prices. The original price will be displayed beside the discounted price. Both price have different styles to be easily recognizable.
 
 Markup:
  <div class="product-thumbnail">
@@ -67,7 +67,7 @@ Pattern 15.2
 /*
 Display a price range
 
-The Product Thumbnail Pattern can also display price ranges. By adding the class .product-price-range to the element with the class .product-price-element the price ranges will be displayed one bellow the other.
+The Product Thumbnail Pattern can also display price ranges. By adding the class .product-price-range to the element with the class .product-price-element the price ranges will be displayed one below the other.
 
 Markup:
  <div class="product-thumbnail">
--- a/src/less/techne-patterns/_product-thumbnail.less
+++ b/src/less/techne-patterns/_product-thumbnail.less
@@ -51,7 +51,7 @@ Markup:
              <span class="product-price-element product-price-original">
                 <span class="product-price-amount">47.48</span>
              </span>
-             <span class="product-price-separator">&nbsp;&mdash;&nbsp;</span>
+             <span class="product-price-separator">&nbsp;&ndash;&nbsp;</span>
              <span class="product-price-element product-price-discounted">
                 <span class="product-price-amount">40.20</span>
                 <span class="product-price-currency">USD</span>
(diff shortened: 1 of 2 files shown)


xak:
@joseegm This solution will force all prices to display on two lines, even if there is not a range of prices. Don't we need to support both options? I recommend the starter page also be updated with the new pattern.

src/less/techne-patterns/_product-thumbnail.less
```diff
@@ -1,28 +1,34 @@
...
+         </div>
+         <div class="product-code">
+             000123456789000
+         </div>
+         <div class="product-price ">
+             <span class="product-price-element product-price-original">
+                <span class="product-price-amount product-price-range">47.48 - 49.48</span>
+             </span>
+             <span class="product-price-element product-price-discounted">
+                <span clipPath="product-price-amount product-price-range">40.20 - 42.48</span>
```
xak: What is `clipPath` here?

src/less/techne-patterns/_product-thumbnail.less
```diff
@@ -114,11 +120,22 @@
     .product-price {
         font-size: 15px;
         margin-bottom: 10px;
-    }
 
-    .product-price-discount {
-        .text-danger;
-        display: inline-block;
+        .product-price-element { 
+            display: block;
```
xak: This breaks the existing single price display. The discounted price will always display below.

**Range**
```
47.48 - 49.48
40.20 - 42.48 USD
```

**One price**
```
47.48
40.20 USD
```

src/less/techne-patterns/_product-thumbnail.less
joseegm: Sloppy autocomplete :)

src/less/techne-patterns/_product-thumbnail.less
joseegm: 👍🏽 I'll consult with Jeanne but I guess. We wanna support both.

xak:
This solution seems a little too complicated for the request. It introduces a lot of new classes and logic requirements in order to output the proper format. Assume a data structure like this ...

```
{
    "prices": [
        { 
            "price": 40.00
        },
        { 
            "price": 40.00,
            "discount": 0.20
        },
        { 
            "price": 40.00,
            "maxprice" 45.00
        },
        { 
            "price": 40.00,
            "maxprice" 45.00,
            "discount": 0.20
        }    
    ]
}
```

We would expect each of these prices entries to output each of the four formats for discounts. The dev would need to write a lot of logic to detect it's a range, detect and compute the discount, add the separator `span`, change the parent container of the currency string, and manage three different modifier classes. This is a lot to manage and prone to errors.

Ideally, each option should be as simple to output as `{% if discount %}<span class="product-price-discounted">{{ computeDiscountedPrice(price, discount) }}{% if maxprice %}–{{ computeDiscountedPrice(maxprice, discount) }}{% endif %}</span>{% endif %}` and be appended to the default option markup.

src/less/techne-patterns/_product-thumbnail.less
```diff
@@ -1,28 +1,112 @@
...
+     </div>
+ </div>
+
+Pattern 15.1
+*/
+
+/*
+Display discounted pices
+
+The Product Tumbnail can also display disconted prices. The original price will be displayed besides the discounted price. Both price have different styles to be easily recognizable.
```
xak: Display discounted prices

The Product Thumbnail can also display discounted prices. The original price will be displayed beside the discounted price. Both price have different styles to be easily recognizable.

src/less/techne-patterns/_product-thumbnail.less
```diff
@@ -1,28 +1,112 @@
...
+     </div>
+ </div>
+
+Pattern 15.2
+*/
+
+/*
+Display a price range
+
+The Product Thumbnail Pattern can also display price ranges. By adding the class .product-price-range to the element with the class .product-price-element the price ranges will be displayed one bellow the other.
```
xak: Display a price range
 
The Product Thumbnail Pattern can also display price ranges. By adding the class `product-price-range` to the element with the class `product-price-element` the price ranges will be displayed one below the other.

src/less/techne-patterns/_product-thumbnail.less
```diff
@@ -1,28 +1,111 @@
...
+         <div class="product-code">
+             000123456789000
+         </div>
+         <div class="product-price ">
+             <span class="product-price-element product-price-original">
+                <span class="product-price-amount">47.48</span></span>
+             <span class="product-price-separator">&nbsp;</span>
+             <span class="product-price-element product-price-effective">
+                <span class="product-price-amount">40.20</span>
+                <span class="product-price-currency">USD</span>
```
xak: can `product-price-currency` go outside and after of the `product-price-element` container(s)? that way it will need no addt'l presentation logic if it just always comes after the price containers.

PR #571, opened 2017-02-21
CCE-640 adds browsersync for component css

xak:


xak:
This adds BrowserSync – https://www.browsersync.io/docs/gulp – to the project, however, it currently only works with component CSS. 

This means if you change a LESS file, the `less` task will run to compile `techne.min.css` and a new task will copy it to the correct KSS directory. The browser will reload to show your change.

The KSS HTML will also recompile and update behind the scenes but you will need to reload the page manually to see any changes you made to the markdown in the LESS files.

I believe there is a problem with watching the same files (LESS has MD and CSS) but then trying to branch the tasks and reload. Right now, unfortunately, the KSS and deploy tasks for putting files where they need to go are poorly organized. Further investigation will be required to enable BrowserSync for all types of changes, i.e., LESS, MD, JS, HTML.

PR #572, opened 2017-02-21
CCE-659 unsticks footer

xak:


xak:
@joseegm The sidebar is fixed now.

joseegm:
@xak tested it again and it looks fine now. Thanks.

docs/template/public/less/_side-nav.less
```diff
@@ -151,26 +151,31 @@ body.has-left-nav .sidebar-nav {
     body.has-left-nav #kss-wrapper {
         padding-left: 265px;
     }
+    .main-content {
+        position: relative; //this is not ideal
```
like-me1: - delete the comment not related to the functionality explanation 
- probably it's from the past here, but would be great to refactor the file according to Less - cascading. Now it feels like pure CSS (e.g. body.has-left-nav repeated multiple times)

xak: I removed the comment. The CSS format was already that way. I suspect it was copied from the KSS default styles and modified. Thanks!

PR #573, opened 2017-02-23
CCE-660, final working crop image functionality with refactored class…

like-me1:
don't forget to do bower update and re-run gulp tasks


src/less/techne-patterns/_crop-img.less
```diff
@@ -0,0 +1,72 @@
...
+      background: #c5cdd6;
+    }
+    .jcrop-vline {
+      width: 3px !important;
+    }
+    .jcrop-hline {
+      height: 3px !important;
+    }
+  }
+}
```
saadmhybris: This whole file needs to be indented with 4 spaces as our code standard. Everything else looks good. We can merge after the indentation is fixed.

PR #574, opened 2017-02-27
CCE-668: adding back lodash.

saadmhybris:
After some quick debugging, looks like we need lodash so adding it back in this PR so that we are not blocked from working on other tickets. We may need a separate ticket to look at lodash and whether or not to include it in the future.


saadmhybris:
ticket - https://jira.hybris.com/browse/CCE-668

PR #650, opened 2017-04-26
Feature/cce 636

saadmhybris:


joseegm:
@saadmhybris This needs to be merged to MASTER so it get pushed to NPM repo, right?

saadmhybris:
@joseegm 

Nope, it's not automated and I feel that it should not be automated. Publishing a new version to NPM is super simple and needs to be a manual step in the release process so that we are consciously deploying as there is no way to rollback the versions from NPM registry (you can unpublish a package but cannot use the same version number again).

PR #654, opened 2017-04-28
CCE-687: Change the feedback basic URL to the new endpoint

joseegm:


docs/template/elements.html
```diff
@@ -283,7 +283,7 @@
 </div>
 
 <!-- feedback button -->
-<div class="feedback-menu feedback-menu--techne feedback-affix"  style="display: none;">
+<div class="feedback-menu feedback-menu--techne feedback-affix">
```
saadmhybris: I think we need to hide this by default. if removing inline style, then we need to hide it in the CSS. 

The reason is because we are checking for a 200 (success) response from the service and then showing the feature. otherwise, the feature would not show if the service is down. Part of Wookies' resilient UI philosophy

Issue #659, opened 2017-05-05
Remove code duplication via CSS cascading


joseegm:
issue goal: remove the code duplication present on https://github.com/SAP/techne/blob/v2.0/dist/css/techne-all.css

On the final generated CSS we distribute we have code duplication. Examples:
https://github.com/SAP/techne/blob/v2.0/dist/css/techne-all.css#L186

Code Extract
```css
.tn-primary-nav__directory {
  display: inline-block;
  margin-bottom: 0;
  padding: 0 20px 0 0; }

.tn-primary-nav__index {
  display: inline-block;
  margin-bottom: 0;
  padding: 0 20px;
  vertical-align: top; }

...
```

Line [186](https://github.com/SAP/techne/blob/v2.0/dist/css/techne-all.css#L186) and [187](https://github.com/SAP/techne/blob/v2.0/dist/css/techne-all.css#L186) are duplicated on the following styles declarations.

This result on unused CSS code been downloaded and parsed.

### Solution

Use the cascade by separating in main and secondary classes 
```ccs 
.tn-primary-nav .tn-primary-nav-index
```

xak:
I'm not entirely sure what you are saying. The example is about duplicated code but the solution is about nesting selectors. I'll address both ...

If you look at the source https://github.com/SAP/techne/blob/v2.0/src/styles/components/primary-nav.scss#L31 you'll see a logical breakdown of the elements under the block. Yes, there could be some improvement and consolidation like ...

```
&__directory, &__index, &__list, &__listitem {
	display: inline-block;
	margin-bottom: 0;
}	
&__directory {
	padding: 0 20px 0 0;
}
&__index {
	padding: 0 20px;	
	vertical-align: top;
}
&__list {
	[aria-selected=true] {
		font-weight: bold;
	}
}
```

_The code there now was the initial rough draft of the component for the POC. We'll circle back to every component and look for optimizations._

Regarding nesting, BEM does not encourage nesting **AT ALL**. Every CSS rule should be a single class to ensure the same specificity for all components. I'm not so dogmatic about it and think nesting sometimes is useful, but for this example, nesting selectors is not required.

joseegm:
This is not the POC. This is the real deal. Optimisations needs to be done early on, otherwise we'll start accumulating technical debt. 
This issue is not fixed, please don't close it.

As explained on https://github.com/SAP/techne/issues/663#issuecomment-299797599
[...]

PR #667, opened 2017-05-08
issue #659 consolidates some declarations for primary nav elements

xak:


xak:
This it to remove the duplicated declarations for the primary nav elements

saadmhybris:
@xak are we checking in .css files as well? Just checking as we had discussed not checking in any generated files.

xak:
Don't we have to check in the `dist` files for them to be available for the packages? I was thinking about setting it up so that when running locally all files are compiled and served from a `.tmp` directory and only get compiled to `dist` when built for production.

saadmhybris:
This is was the case when we didn't have automatic deployment or when we were not compiling files on the server. Now that this issue was resolved in the dev deployment bamboo plan, I don't think we need to check in compiled files anymore.

xak:
Hmm, isn't the package pulled from Github. There's nothing in the `bower.json` or `package.json` that points to our server location, right? My https://github.com/thestrongelement/strong-css repo is packaged but isn't hosted anywhere other than Github. Maybe I'm wrong though.

xak:
I'll merge this and we can later investigate serving locally from a different directory. This is annoying.

Issue #657, opened 2017-05-03
Ops: Refactor gulp setup


xak:
`gulp-sequence` requires all gulp tasks to `return` their stream so it can execute properly (a good practice in general). That does not allow for bunching tasks and calling them in functions.

```
const task = (cb) => {
	foo()
	bar()
}	
const foo = (cb) => {
 	return gulp.src('./*').pipe(gulp.dest('./'));
}
const bar = (cb) => {
 	return gulp.src('./*').pipe(gulp.dest('./'));
}
gulp.task('task', task)
```

The JS files need to be refactored so each file has only one gulp task. This looks like a good model ... https://medium.com/@_rywar/spreading-gulp-tasks-into-multiple-files-2f63d8c959d5

xak:
@saadmhybris I created a new issue to capture this but you can implement as part of #638. No big deal.

PR #674, opened 2017-05-12
Feature/cce 643 css optimization

like-me1:


xak:
Optimization and minification has been incorporated into gulp refactoring #657

Issue #681, opened 2017-05-16
Consolidate build commands


joseegm:
Create a new `gulp build` to build the library and the website in one simple command. Executing relevant build tasks as described on https://github.com/SAP/techne/blob/v2.0/ops/README.md

### Usage

`gulp build` build the library and the website for development.

`gulp build --production` build the library and website for production.

joseegm:
Also add a main clean tasks `gulp clean`.

Issue #647, opened 2017-04-21
Docs: Website deployment task


xak:
The current `v2.0` branch has a build task for the docs — `gulp docs-build`

Investigate if we need additional task(s) that will be run by Bamboo for various environments. A `--production` (or environment) flag may be employed to enable additional minification and optimizations if applicable. 

The docs build task currently calls `pkg-build` first which works OK for local development but is not ideal. We should consider decoupling these. Watch tasks handle compilation for development purposes and the `default` sequence can be changed to run `pkg-build` first.

Consider pulling Techne from `bower` in the deployment scenario. The local build process compiles the library in the project and copies it to `www/css` directory. This could introduce differences in the website version of Techne vs. the published packaged version. 

For each numbered releases, this would necessitate a publishing order:
1. Publish the package
2. Bump the Techne bower version in the project
3. Deploy the site

xak:
Related rot #617

joseegm:
@xak You can use this one now. It was ready a month ago, merged and unmerged. 
issue https://github.com/SAP/techne/issues/681, pr https://github.com/SAP/techne/pull/682

xak:
# Blocked by #617

saadmhybris:
no longer valid.

PR #682, opened 2017-05-16
#681: Global Build and Clean gulp commands added. Docs clean refactor…

joseegm:
…ed to use the data on the config file

### DIFF
--- a/ops/README.md
+++ b/ops/README.md
@@ -61,3 +61,10 @@ Local development requires a server and includes watch tasks which auto-compiles
 ### Individual tasks
 * `gulp dev-serve` initializes BrowserSync and watches `www` for changes
 * `gulp dev-watch` kicks off compilation tasks when `src` files change
+
+
+### Short-hand commands
+* `gulp build` build the package and the docs.
+* `gulp build --production` build the package and the docs for production.
+
+* `gulp clean` clean all destination folders used generated with `gulp build`. It also accept the parameter `--production` to clean the production folders.
--- a/ops/gulp/build.js
+++ b/ops/gulp/build.js
@@ -1,0 +1,3 @@
+const gulp = require('gulp');
+
+gulp.task('build', ['pkg-build','docs-build']);
--- a/ops/gulp/clean.js
+++ b/ops/gulp/clean.js
@@ -1,0 +1,3 @@
+const gulp   = require('gulp')
+
+gulp.task('clean', ['pkg-clean', 'docs-clean']);
--- a/ops/gulp/docs-clean.js
+++ b/ops/gulp/docs-clean.js
@@ -1,8 +1,10 @@
+const config = require('../config');
+
 const gulp = require('gulp');
 const del = require('del');
 
 const task = (cb) => {
-    return del(['./www/**/*']);
+    return del([config.root.www]);
 }
 
 gulp.task('docs-clean', task);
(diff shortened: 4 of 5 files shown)


xak:
These are not needed at this time. Deployment and publishing tasks where both outputs will be built and coordinated have yet to be written. I prefer to maintain a separation between the site and the package until we have an explicit need to do otherwise.

Running `gulp docs-build` builds the package first and accomplishes the same thing. 

However, this should be revisited since there are times when we deploy the site independently of a package version bump, and coupling the two risks introducing unintended library changes. Of course, it's not hard to manage this manually for hot fixes, but we should look at this as part of the deployment implementation if we want to build in concrete separation. For example, we could rely on bower to pull the latest version of Techne when production builds are executed by Bamboo. See #647

joseegm:
@xak Think I didn't explained the rationale of this PR, some more explanation is needed.

Since we are using unconventional gulp tasks commands, see other libraries for reference:
https://github.com/angular/material2/blob/master/tools/gulp/tasks/publish.ts#L104

Convention in the angular world -our main users front-end framework- is `gulp <action>:<environment>`, with common action names like `gulp build` to build the whole solution, `gulp build:dist` for production.

Our current implementation uses `gulp <module>-<action> --<environment>` that's why we need a conventional way to build the whole package with easy, using the short commands `gulp build` and `gulp build --production` instead of `gulp docs-build` which is something different we have to remember and doesn't reflect if it builds the whole solution, only the docs.

Now you have more elements.

xak:
@joseegm This merge has been reverted. It was not your call to merge this. I indicated that this was not needed at this time therefore it was not merged. There are three developers working the project framework and when this is needed it will be added at that time. At this time, it only adds clutter to the project.

If you have a tech-related issue, please create the issue and leave it in the backlog. It will be prioritized along with every other issue. You do not need to do the task yourself or add it to the implementation board with your own priority.

Thanks.

joseegm:
@xak 

The comment https://github.com/SAP/techne/pull/682#issuecomment-302008386 was sitting without answer for a very long while. 

As per info on that comment major editors and IDEs pick **conventional** gulp tasks out of the box, they usually named `build`. See:
* https://github.com/AtomBuild/atom-build-gulp/blob/master/spec/build-gulp-spec.js#L65
* https://code.visualstudio.com/docs/editor/tasks
* https://blogs.msdn.microsoft.com/webdev/2016/01/06/task-runners-in-visual-studio-2015/

Which makes building the library or website with a short-cut a breeze. Not like right now that we don't have **conventional named** gulp tasks.

The use case is valid and meaningful, fully documented on #681, the code was valid, no conflict on the PR... 

Do you suggest drop that and open a ticket to to the same thing? 
To develop again the same thing?

Let me know if you need me to open a Technical ticket for this, so it gets developed again.

xak:
@joseegm The examples you include do not assume any conventional task names. They either pick up the tasks as defined in the gulpfile or require a separate tasks definition file. I'm not debating that many projects use `build`, as I've said before, we may need this in the future but it is not an MVP need now.

With this and any tech-related issue or concerns, you should open an issue and explain why you are requesting the change. That's all you have to do. If the issue is important then let me know and we can discuss the priority, otherwise the issue will be assessed and prioritized along with everything else. For this, you created the issue, added it to the implementation board, coded it, moved the ticket to done, and created a PR — all within a few hours. That is a hotfix process, not a feature process. 

If you want to have a closer look at the tasks in use and recommend a new approach that is more in line with other projects, then open a new issue and outline your approach. We'll take a look and validate it against all of our needs.

joseegm:
@xak All right, then I'll open an issue and request the same change it is already developed. Good.

PR #684, opened 2017-05-19
Feature/646

saadmhybris:


src/icons/css/sprite.css
```diff
@@ -0,0 +1,506 @@
...
+    content:' ';
+    vertical-align:middle;
+    display: inline-block;
+    background-image: url("../svg/sprite.svg");
+    background-repeat: no-repeat;
+    background-size: 13.2em 712.8em;
+}
+
+.no-svg .hyicon:before {
+    background-image: url("../svg/sprite.png");
```
xak: No PNGs are being output. Do we even need the PNGs? If not, these fallback rules should be removed.

joseegm:
Icons already have a place under "Guidance & Indicators" 
https://github.com/SAP/techne/wiki/Techne-Components

src/icons/css/sprite.css
joseegm: Nope. We don't need PNGs, we don't have them now either.

ops/gulp/icons.js
```diff
@@ -0,0 +1,17 @@
...
+    .pipe(svgmin())
+    .pipe(svgSprite({
+        selector: "hyicon-%f",
+        common: "hyicon",
+        templates: { scss: true }
+    }))
+    .pipe(gulp.dest("./src/icons"));
+}
+
+gulp.task('hyicon', task);
```
saadmhybris: done

src/icons/css/sprite.css
saadmhybris: Currently, there is no option to disable the generation of this class. I have opened the issue with the plugin author git repo. We may have to just live with it for now.

saadmhybris: sovled it by using a custom template.

saadmhybris:
@xak can you please take another look at this when you get a chance. the icons component page (http://localhost:3000/icons.html) is a rough draft to just to make sure icons are generating/rendering ok. there are several changes after merge so I am expecting some more work on this before it's merged.

ops/gulp/icons-build.js
```diff
@@ -0,0 +1,44 @@
...
+const config = {
+    templates: { css: require("fs").readFileSync("./src/icons/icons-template.css", "utf-8") },
+    selector: "hyicon-%f",
+    common: "hyicon",
+    svg: { sprite: "css/icons.svg" },
+    cssFile: "css/icons.css",
+    preview: {sprite: "icons.html"}
+};
+
+// generate icons
```
xak: This file doesn't exactly match the pattern for the other build files.

This should be the default task to build the icons where you sequence the individual tasks together — taking in the `src` and doing all the magic you need to do to output the files.  I would expect to run `gulp icons-build` to generate the icons.

Each of these tasks should be separated like`icons-generate`, `icons-copy`, etc. It may seem like overkill but it keeps this file clean as the main roadmap to understand exactly what is happening.

ops/gulp/icons-build.js
```diff
@@ -0,0 +1,44 @@
...
+           .pipe(gulp.dest("./src/icons"))
+}
+
+gulp.task('generateIcons', generateIcons);
+module.exports = generateIcons;
+
+// copy svg file 
+const copyIconsSvg = (cb) => {
+    return gulp.src('./src/icons/css/*.svg')
+           .pipe(gulp.dest("./www/css"));
```
xak: This task is related to the docs. It would be better to have a `docs-icons` task to grab what is needed from the `src` directory when the `docs` are built.

ops/gulp/pkg-icons.js
```diff
@@ -0,0 +1,9 @@
+const gulp = require('gulp');
+const gulpSequence = require('gulp-sequence');
+
+const task = (cb) => {
+    gulpSequence('generateIcons','copyIconsSvg', 'copyIconsCss', cb);
```
xak: This task should assume that the icons files have already been built. It's job should be only related to whatever needs to happen when the package is being built and moving files from `src` to `dist`.

ops/gulp/docs-build.js
```diff
@@ -5,7 +5,7 @@ let environment = require('../lib/environment');
 
 const task = (cb) => {
 
-    gulpSequence('pkg-build', 'docs-clean', ['docs-resources', 'docs-html', 'docs-css', 'docs-styleguide', 'docs-js'], cb);
+    gulpSequence('pkg-build', 'docs-clean', ['docs-resources', 'docs-html', 'docs-css', 'docs-styleguide', 'docs-js', 'pkg-icons'], cb);
```
xak: `pkg-icons` shouldn't be called separately. It should be part of the `pkg-build` process that runs before any docs tasks even begin. As mentioned below, if you need a `docs-icons` task to reach into `src` and grab the icons sprite then create a separate one for just that.

src/icons/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Icons are generated using SVG files in the *raw-svg* folder and converting them into a SVG sprite found under svg/sprite.svg. Required gulp plugins are `gulp-svg-sprites` for sprite generation and `gulp-svgmin` for cleaning and optimizing SVG files. We are no longer using icon fonts. 
+
+## Generated Files 
+
+The generated files include `css folder`, `svg folder` and `icons-preview.html`. Please do not modify these files as any changes made here will be erased once the gulp task for generating icons is executed.
+
+## Gulp Task 
+
+Gulp task for generating icons is under `ops/gulp/icons.js`. To generate icons, use the command `gulp icons`. The task will generate `icons.svg`, `icons.css`, and `icons.html`
```
xak: This doesn't seem right, I don't see how to generate the icons independently.

ops/gulp/dev-watch.js
```diff
@@ -13,7 +13,10 @@ const task = () => {
 	gulp.watch([`./docs/**/*.html`,`./src/**/*.html`], ['docs-html']);
 
 	//watch component styles
-	gulp.watch(`./src/**/*.scss`, ['pkg-css']);
+	gulp.watch([`./src/**/*.scss`], ['pkg-css']);
+
+	//watch icons 
+	gulp.watch([`./src/icons/**/*.*`], ['pkg-icons']);
```
xak: Not much harm here, but I don't think we actually need to watch this. Adding new icons should be rare enough that we could explicitly run `gulp icons-build` but regardless this should kick off `icons-build` at which point a new `icons.scss` will get added to the `src/styles` directory and be picked up by the `pkg-css` task.

xak:
Note, the icons.html page works just fine.

saadmhybris:
@xak - Please have another look. I have refactored all the gulp tasks as you described.

ops/gulp/icons-generate.js
```diff
@@ -0,0 +1,24 @@
...
+const svgSprite = require("gulp-svg-sprites");
+
+// svgSprite configuration
+const config = {
+    templates: { css: require("fs").readFileSync("./src/icons/techne-icons-template.css", "utf-8") },
+    selector: "hyicon-%f",
+    common: "hyicon",
+    svg: { sprite: "css/techne-icons.svg" },
+    cssFile: "css/techne-icons.css",
+    preview: {sprite: "techne-icons.html"}
```
xak: Keep the names for the generated file simple, no `techne-` prefix. Then do the prefixing in the `pkg` tasks along with the other files. So if we decide to change to `tn-all.css` it'll be easier to change that as part of the packaging tasks instead of having to dip into this file.

ops/gulp/docs-icons.js
```diff
@@ -0,0 +1,9 @@
+const gulp = require('gulp');
+
+const task = (cb) => {
+    return gulp.src('./src/styles/icons/**.svg')
```
xak: Pull this from the `dist` directory instead of `src`. For the most part, the docs shouldn't need much from the `src` directory, the main exception being the `src/templates/` includes used for the component pages.

Just like the `icons-build` is the pre-build step before the package is built, you should assume that the package is built and all the files required for the website are available in `dist` (or `tmp`) before the site Is built. The order must be 1) build icons, 2) build package, 3) build site.

This ensures there is no disconnect between the `src` and the `dist` versions. Check out the `docs-css` task.

saadmhybris:
@xak  - please have a look when you get a change... third time is a charm *fingers crossed*

joseegm:
@xak 
Are the changes of https://github.com/SAP/techne/pull/684/commits/2c574c57e724cde7cb210a0fbdab97923b8c2780 already reviewed?

xak:
@saadmhybris I went ahead and made a few updates. For example, into of building the html into `src/styles` I moved it to `src/templates` and added custom template to output the example file you had included. So `localhost:3000/icons.html` pulls the generated template now.

PR #686, opened 2017-05-22
Feature/cce 639 implement scripts

like-me1:
gulp task for compiling JS files is created. It works fine with gulp docs-js, and gulp docs-js --production and creates techne-all.js and techne-all.min.js respectively. The only one thing is that with npm start --production the environment.production variable set to false therefore I used gulp docs-js --production to test the gulp task.

### DIFF
--- a/docs/styleguide/templates/master.html
+++ b/docs/styleguide/templates/master.html
@@ -8,10 +8,6 @@
 {% block css %}
 
 {% endblock %}
-      <script src="js/techne-{% if debug %}debug{% else %}all{% endif %}{% if env == 'production' %}.min{% endif %}.js"></script>
-{% block js %}
-
-{% endblock %}
   </head>
   <body id="app--{{ app.id }}">
   <!--
@@ -64,5 +60,9 @@ _gaq.push(['_setAccount', 'UA-00000000-0']); _gaq.push(['_trackPageview']);
 })();
     </script>
 {% endif %}
+<script src="js/techne-styleguide{% if env == 'production' %}.min{% endif %}.js"></script>
+{% block js %}
+
+{% endblock %}
   </body>
 </html>
--- a/docs/styleguide/templates/master.html
+++ b/docs/styleguide/templates/master.html
@@ -2,10 +2,14 @@
 <html lang="en-us">
   <head>
 {% include "app_meta.html" %}
-	<link rel="stylesheet" href="css/techne-{% if debug %}debug{% else %}all{% endif %}{% if env == "production" %}.min{% endif %}.css">
+	<link rel="stylesheet" href="css/techne-{% if debug %}debug{% else %}all{% endif %}{% if env == 'production' %}.min{% endif %}.css">
 
     <link rel="stylesheet" href="css/techne-styleguide.css">
 {% block css %}
+
+{% endblock %}
+      <script src="js/techne-{% if debug %}debug{% else %}all{% endif %}{% if env == 'production' %}.min{% endif %}.js"></script>
+{% block js %}
 
 {% endblock %}
   </head>
--- a/ops/gulp/dev-watch.js
+++ b/ops/gulp/dev-watch.js
@@ -21,6 +21,9 @@ const task = () => {
 	//update styleguide styles
 	gulp.watch([`./docs/styleguide/styles/*.scss`], ['docs-styleguide']);
 
+	//watch docs js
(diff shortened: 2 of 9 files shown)


docs/styleguide/templates/master.html
```diff
@@ -2,12 +2,16 @@
...
   <head>
 {% include "app_meta.html" %}
-	<link rel="stylesheet" href="css/techne-{% if debug %}debug{% else %}all{% endif %}{% if env == "production" %}.min{% endif %}.css">
+	<link rel="stylesheet" href="css/techne-{% if debug %}debug{% else %}all{% endif %}{% if env == 'production' %}.min{% endif %}.css">
 
     <link rel="stylesheet" href="css/techne-styleguide.css">
 {% block css %}
 
 {% endblock %}
+      <script src="js/techne-{% if debug %}debug{% else %}all{% endif %}{% if env == 'production' %}.min{% endif %}.js"></script>
```
xak: A few things ...

1. Since this JS is specific to the style guide and not the library, let's call it `techne-styleguide.js` to match the name of the styleguide CSS.
2. All of this should go at the bottom of the HTML page, before the closing `<body>` tag.
3. The debug logic can be removed since there is no debug JS.

ops/gulp/dev-watch.js
```diff
@@ -21,6 +21,9 @@ const task = () => {
 	//update styleguide styles
 	gulp.watch([`./docs/styleguide/styles/*.scss`], ['docs-styleguide']);
 
+	//watch docs js
+	gulp.watch([`./src/scripts/*.js`,`./ops/gulp/*.js`], ['docs-js']);
```
xak: No need to watch `./ops/gulp/*.js`. This would run `docs-js` when any ops JS file is modified and most of the JS files are unrelated to the site JS.

Also, the watch task here will compile the JS but will not update the browser automatically. See `dev-serve.js` where Browsersync file watching is specified to inject the JS into the page when running the local server.

ops/gulp/docs-js.js
```diff
@@ -0,0 +1,27 @@
...
+const gp_concat = require('gulp-concat');
+const gp_rename = require('gulp-rename');
+const gp_uglify = require('gulp-uglify');
+const gulpif = require('gulp-if');
+const gulp_jslint = require('gulp-jslint');
+
+let environment = require('../lib/environment');
+
+const paths = {
+    src: './src/scripts',
```
xak: Since this JS is for the site only and not part of the library, it should be in `docs` and not `src`.

ops/gulp/docs-js.js
```diff
@@ -0,0 +1,27 @@
...
+    dest: './www/js'
+}
+
+let techneDestJS = environment.production ? 'techne-all.min.js' : 'techne-all.js';
+
+const task = (cb) => {
+    return gulp.src([`${paths.src}/*`])
+        .pipe(gulp_jslint())
```
xak: Does JS lint print errors to the console?

ops/gulp/docs-js.js
```diff
@@ -0,0 +1,27 @@
...
+        .pipe(gp_concat('concat.js'))
```
xak: I'm not sure concatenation will work here. This will take all the files in the directory and concat them in alpha order, right? This may not be a problem depending on how we organize the files but we may need to control the order of the files, for example, if we are including jQuery here it would have to be included first.

docs/styleguide/templates/master.html
like-me1: @xak done

ops/gulp/dev-watch.js
like-me1: @xak deleted ./ops/gulp/*.js from docs-js and added watcher for JS-files in dev-serve.js

ops/gulp/docs-js.js
like-me1: done

ops/gulp/docs-js.js
like-me1: nope, used gulp-eslint instead

ops/gulp/docs-js.js
like-me1: when using gulp-concat files will be concatenated in the order that they are specified in the gulp.src function.If we need some custom includes we can specify it in gulp.src or use https://www.npmjs.com/package/gulp-order. But for now I'd leave it as it is. For now I can't oversee the future exceptions.

like-me1:
@xak ready for check

xak:
:shipit:

Issue #683, opened 2017-05-17
Pkg: Implement standard component contribution process


xak:
Build out and document the process for adding a new component to the package and the website.

- SASS
- Data
- Template

Issue #663, opened 2017-05-05
Remove expensive CSS query


joseegm:
issue goal: remove expensive CSS queries

On 
https://github.com/SAP/techne/blob/v2.0/dist/css/techne-all.css#L165
the CSS query 
```css
[class*="tn-primary-nav--"] { }
```
That's a expensive query to perform and is equivalent to 
```css
.tn-primary-nav { }
```
which is the default query for CSS, performance on browser implementations is optimised for it. Then use the cascade as described on #659

xak:
There is an important difference in the two examples. Back to BEM, this shows a modifier for the component ...

```
.foo {
	border: solid 1px black;
	background-color: white;
}
.foo--highlighted {
	background-color: yellow;
}
```

In the markup, BEM notation (like Bootstrap) requires duplicate classes where `foo` gives you the base styles and the modifier class gives you the modifications ...

```
<div class="foo foo--highlighted">
</div>
```

Adding the attribute selector means there is no need for the duplication in the markup ...

```
[class*="foo--"], .foo {
	border: solid 1px black;
	background-color: white;
}
```

If you look at the source you'll see that this was testing a concept. https://github.com/SAP/techne/blob/v2.0/src/styles/components/primary-nav.scss#L26 

Not much attention has been given yet to component SASS structure and coding recommendations. As more components are built out we can assess the value. The attribute selector may not be useful or too hard to understand for some devs but I've found nothing to indicate this an expensive selector for performance being no worse than using multiple child selectors.

joseegm:
This is issue is not solved. Please don't close it.

Bootstrap is not using BEM 
https://github.com/twbs/bootstrap/blob/v4-dev/dist/css/bootstrap.css#L4055
and Material2 is not using it either
https://github.com/angular/material2/blob/master/src/lib/card/card.scss

It is not a easy to type class names for developers and it creates a duplicated code as explained on #659 

On a 1x1 benchmark it is more expensive
https://benfrain.com/css-performance-revisited-selectors-bloat-expensive-styles/
Comparing results of 10 and 20 tests

We have to ditch BEM notation since it doesn't bring any real value and create duplicate code

xak:
I'm fine removing the attribute selector. There are other ways to achieve the same thing in CSS. 

But I strongly disagree about BEM. It brings **A LOT** of value and unless we align around another approach or I see a specific custom implementation recommendation, there is no reason to change our approach. 

[...]

PR #690, opened 2017-05-25
issue #683 build out some more of the site structure; implements some…

xak:
… repeatable patterns for creating components


xak:
@saadmhybris There's quite a bit here.

1. Refactors some of the app-level containers and adds very high-level styles so we can begin building out the site CSS esp. the top and sidenavs.
2. Moves primary nav out of the library and to docs since it won't be a component. I also assume based on Leo's direction that we will be using the current Yaas.io nav.
3. Builds out the buttons and tables DOCS pages using a repeating pattern. Sets the base component and then loops through modifiers. This may or may not hold up which is why I'm going through the specs for each component. See the Wiki.
4. Uses the Nunjucks macros functionality to create actual components that accept arguments to render themselves. I create in `src/templates` group files like `actions.html` that can hold all those components. These are imported into the DOCS page where arguments are defined and passed like `{{ table(header=component.data.header, rows=component.data.rows) }}`
5. Other kinds of reorganization.

Thanks.

src/styles/components/table.scss
```diff
@@ -0,0 +1,64 @@
...
+    @include reset;
+    border-collapse: collapse;
+    width: 100%;
+    background-color: color(light);
+
+    //VARIANTS
+    //each variant MUST extend base block
+    // &--modifier {
+    //     @extend .#{$block};
+    // }
```
saadmhybris: do we need to retain this commented code? in the past, we did not keep commented to the keep the files clean as you could recover the code from file history

xak: This is here as an example to show how a dev would need to extend the block when adding a modifier. I can remove it and create an example SCSS file for reference.

NOTE: This is only necessary to avoid duplicating classes like `<table class="tn-table tn-table--foo">`. If we want to build the library to support  `<table class="tn-table--foo">` where the modified class get all the base rules , then we have to code it like the example. My alternative approach was squashed in #663.

saadmhybris: ok makes sense. maybe we can add some more comments to indicate that it's example code.

PR #706, opened 2017-06-07
CCE-692 upload options small icon for smartedit

SAPMontreal:


xak:
@saad This needs to be added to the 2.0 icon set

Issue #712, opened 2017-06-12
Pkg: Update color palette


xak:
@LeoT7508 provided updated color palette.

![screen shot 2017-06-12 at 9 08 54 am](https://user-images.githubusercontent.com/424119/27038214-f2024ffc-4f4f-11e7-9600-be73d496792f.png)

Need to update library variables.

xak:
Also updated Invision reference sheet — https://sapse.invisionapp.com/share/DTBXJCSJW#/235917132_Core-Colors

joseegm:
I used this method to create a coherent palette from a "default" and "accent" colors.

from https://github.com/SAP/techne/blob/feature/CCE-v2.0/src/techne/system/colors.scss
```scss
$tn-color-default: #006fbb !default;
$tn-color-default-light: lighten($tn-color-default, 10%);
$tn-color-default-dark: darken($tn-color-default, 10%);

$tn-color-accent: #F0AB00 !default;
$tn-color-accent-light: lighten($tn-color-accent, 10%);
$tn-color-accent-dark: darken($tn-color-accent, 10%);

$tn-color-neutral: #21262c !default;
$tn-color-neutral-lightest: lighten($tn-color-neutral, 81%);
$tn-color-neutral-lighter: lighten($tn-color-neutral, 70%);
$tn-color-neutral-light: lighten($tn-color-neutral, 40%);
$tn-color-neutral-dark: darken($tn-color-neutral, 10%);

$tn-color-positive: #18C407;
.tn-color-positive { color: $tn-color-positive; }
```
*Note*: Ignore the names above, they can be better.

Is a simple system, every color have a "normal", "lighter" and "darker" shade. 
The grays are a different story, more shades are needed for that color.

@xak Do we need a separate issue for this?

xak:
@joseegm This is a design system decision, if @LeoT7508 needs four variations of each color then we can implement this system. I haven't seen the need for this yet, however.

I would note that this works great as long as the colors are carried through the design system and easy to "pick out" of a design. From a developer standpoint, it's a bit more difficult since Photoshop, Sketch nor XD "support" this type of color definition ... it's just HEX or RBG.

[...]

PR #713, opened 2017-06-12
#712 updates color palette

xak:


xak:
@like-me1 Could you verify that the hex values are correct related to the colors Leo provided at https://github.com/SAP/techne/issues/712?

Most of the other changes are related to changing the former "action" to "accent" and changing the former "link" to "action" and updating the global variable definitions.

Thanks.

xak:
Merging this. I need these now. :)

Issue #709, opened 2017-06-09
Pkg: Build Cards component


xak:
# Inputs
**Design** 
Background at #678 
https://app.box.com/folder/29116880878

**Prototype** 
https://sapse.invisionapp.com/share/DTBXJCSJW#/239183941_Techne_2-0_Cards_CommonCards
https://sapse.invisionapp.com/share/DTBXJCSJW#/239183942_Techne_2-0_Cards_with_Supporting_Image

## Basic block structure
```
.tn-card+(--static)+(.is-disabled|[aria-disabled=true])
    .tn-card__media?
    .tn-card__content
        .tn-card__header
        .tn-card__description
        .tn-card__label?
```
See discussion at #692

# Outputs
- Card (Common Cards)
- Media cards (Cards with Supporting image)

# Notes
See https://hybris.box.com/s/ojarlct2k6gcp86r24qylqwo7a60xevf

_These are applicable to this issue_
Cards will always only have:
- Primary value (wrap if necessary) 
- Secondary value (one line with “…” never wraps for spacing consistency)
- Status Label (one line, if there are multiple status’ they are separated by a 1px vertical rule,                 wrapping if necessary)

joseegm:
@xak 
This issue needs to labeled accordingly when it is ready to review. 
Which label you are using for that?

joseegm:
### New Design for reference
<img width="1155" alt="screen shot 2017-06-27 at 15 04 05" src="https://user-images.githubusercontent.com/22662903/27588938-14b3d6fa-5b4a-11e7-9281-370ae7f9059d.png">

<img width="1281" alt="screen shot 2017-06-27 at 15 06 10" src="https://user-images.githubusercontent.com/22662903/27589068-5ff85ed8-5b4a-11e7-9167-2c32a2e07ec7.png">

PR #722, opened 2017-06-13
Feature/709

xak:


xak:
For #709

xak:
This adds a new page at `components/card.html`— the link should show in the secondary navigation under "Collections".

The common card types are shown along with hover and disabled states, also shown is an additional card with the left side image. The page itself is not formatted very well but the Cards themselves should match the designs linked in #709

src/styles/components/card.scss
```diff
@@ -0,0 +1,126 @@
...
+@import "../core/functions";
+
+/*!
+    card structure
+    .tn-card+(--static)+(.is-disabled)
+        .tn-card__media+(--fill)
+        .tn-card__content
+            .tn-card__header, .tn-card__description, .tn-card__label
+*/
+$block: ns(card);
```
like-me1: not self-explaining would be great to have some description next to it.

src/styles/components/card.scss
```diff
@@ -0,0 +1,126 @@
...
+            background-size: cover;
+        }
+
+    }
+
+    &__content {
+        padding: ($tn-card-padding - 5px) $tn-card-padding;
+        width: 100%;
+    }
+
```
like-me1: Was double space made on purpose?

src/styles/components/card.scss
```diff
@@ -0,0 +1,126 @@
...
+            }
+            &:last-child::after {
+                content: "";
+                padding-left: 0;
+                padding-right: 0;
+            }
+            &:nth-child(2) {
+                @include weight(semi);
+            }
+
```
like-me1: space?

src/styles/components/card.scss
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
```
like-me1: 3 empty lines

src/styles/components/card.scss
```diff
@@ -0,0 +1,126 @@
...
+    }
+    //STATES *******************************************
+    &[aria-disabled="true"], &.is-disabled {
+        background-color: $tn-card-color-disabled;
+        cursor: default;
+        * {
+            color: color(dark);
+        }
+    }
+
```
like-me1: and finally 4 empty lines? :)

src/styles/helpers/_backgrounds.scss
```diff
@@ -1,17 +1,19 @@
 @import "../core/settings";
 
-.has-background-image {
+$ns: ns();
+
```
like-me1: what is "ns"?

src/styles/helpers/_colors.scss
```diff
@@ -0,0 +1,10 @@
+@import "../core/settings";
+@import "../core/functions";
+
+$ns: ns();
```
like-me1: sorry, this line doesn't say anything to me from the naming prospect of view. Some comments are appreciated

src/styles/components/card.scss
```diff
@@ -0,0 +1,108 @@
...
+@import "../core/functions";
+/*!
+    card structure
+    .tn-card+(--static)+(.is-disabled)
+        .tn-card__media+(--fill)
+        .tn-card__content
+            .tn-card__header, .tn-card__description, .tn-card__label
+*/
+//set component name using namespace function, then use #{$block} below
+$block: ns(card);
```
xak: added comment here and changed the name spacing a bit. see `core/_settings` where namespace is set for the library, then see `core/_functions` for the `ns()` code. For the components, the line here is part of the template where you only have to put the component name here and it will be output as `tn-card`, for example (#683 will automate this). If you don't pass an arg to the function, it will return only `tn-` which is what you see in the helpers. I added some comments to those as well.

xak:
@like-me1 Should be all fixed. 👍

PR #727, opened 2017-06-15
Feature/683

xak:


xak:
The `ops/README` has a good overview of what this does.
https://github.com/SAP/techne/pull/727/files#diff-08305bcb996f50793abd24ca4b1b7812

## NOTE: #722  implements a second-level directory for components (i.e., `components/`). Once that PR is merged, this code will need to be updated to output the docs component HTML into that directory. It's a single line change and I can do it, but I wanted to note it anyway.

Issue #731, opened 2017-06-16
Hotifx: Add temp background to sidebar nav


xak:
#715 removed background for the sidebar so it's basically hidden now. Need to add it back.

Issue #717, opened 2017-06-13
Docs: secondary navigation styling


xak:
# Unblocked by #689

The secondary navigation needs styling to match the rest of yaas.io. The Tools section has a nice implementation — https://devportal.yaas.io/tools/. _You can ignore the collapsible control for now, we can revisit that as an enhancement later._

<img width="364" alt="screen shot 2017-06-13 at 9 23 32 am" src="https://user-images.githubusercontent.com/424119/27087300-4091153e-501a-11e7-938c-f7a3055c6db4.png">

NOTE: Some additional work will be needed in the template code to set the "selected" state for the nav item. That can be done as part of this or separately. See #718

PR #732, opened 2017-06-16
#731 hotfix to add sidebar background

xak:


xak:
This is a demo fix only until sidebar nav is complete #717

Issue #743, opened 2017-06-20
Pkg: Small changes to Cards


xak:
- Change `--static` to -`--readonly` to align with EC
- Better match hover glow to the PSD
- Update disabled background to improve a11y
- Change `__label` element to `__status`

PR #745, opened 2017-06-20
#743 small updates to Cards design styles

xak:


xak:
Small changes based on today's review meeting and from Leo design review.

Issue #691, opened 2017-05-25
Docs: Build out the main content area


xak:
The initial implementation should focus on the "article" top section (page title and description) and the layout containers (spacing and alignment).

# Inputs
**Specs:** https://app.box.com/files/0/f/27696923061
**Invision:** https://sapse.invisionapp.com/share/DTBXJCSJW#/238743129_Docs-Grid and https://sapse.invisionapp.com/share/DTBXJCSJW#/238743128_Docs-MainContent

Use the Grid and Guides pages to build out the main content area and the high-level sections.

# Outputs
* Implement baseline responsive behavior (assume containers stack at smaller sizes)
* Define max content width (we don't want fully fluid behavior soon massive displays)
* Implement grid and layout classes (or mixins) to allow for multiple column containers
* Build page title and description containers to spec (these are fairly global and could be added to the master template if that makes sense)
* Begin implementation of type specs based on header and other content sizes

joseegm:
Design at #694

xak:
Note that this task is not about building out the full spec from #694 but rather the high-level structural containers that each page will inherit.

joseegm:
@xak Okay. Description of the issues wasn't clear enough.

Please add the issue for the pimplementation of the components page. Thanks.

xak:
The components page implementation issue will be created when #694 is complete.

joseegm:
@xak That's a waste of time since the main elements needed to build the markup are already delineated on 
https://github.com/SAP/techne/issues/694

xak:
As I understand it, the process was to work through the UX and design and then move toward implementation. Not all issues have been resolved for #694 . I'm creating implementation tasks only when I'm satisfied they can be completed successfully.

joseegm:
Waste of time on the context of:
https://github.com/SAP/techne/milestone/10

 The main elements are defined on the wire frames already.

PR #747, opened 2017-06-20
#691 adds page-level css, mixins, secondary page template with sidebar

xak:


xak:
This does a fair amount ...

- Adds a new `secondary.html` template to handle the sidebar, page title and intro have moved to that template so they do not have to be coded into every page. 
- Adds `page` and `sections` SCSS files, more variables and mixins
- Styles the page title and intro containers
- Adds `docs-section` containers to the component pages so they have proper side and bottom padding
- Introduces `flow` classes with some basic responsive behavior, see `debug.html` for flow layout examples.

I think that's it other than fixing some indents here and there.

Issue #729, opened 2017-06-15
Pkg: Implement type specs


xak:
# Inputs
See #721 
https://sapse.invisionapp.com/share/DTBXJCSJW#/239182587_Techne_2-0_Typography_Master

# Outputs
Modify [_settings.scss](https://github.com/SAP/techne/blob/v2.0/src/styles/core/_settings.scss#L27)
- 16px is the base
- Sizes should be relative to base
- Line height is relative to size (no unit necessary)

Add new type helpers
- Move helpers in [_general.scss](https://github.com/SAP/techne/blob/v2.0/src/styles/helpers/_general.scss#L38) to `_type.scss`
- `.tn-has-font-plus-SIZE`, `.tn-has-font-minus-SIZE`, `.tn-has-font-family-FAMILY`, etc.

Build spec sheet
- A page for the website to show the usage of helper classes

PR #750, opened 2017-06-21
#729 updates font sizes, enhances mixin, sets helper classes

xak:
Specs have been added to the Styles page at http://localhost:3000/styles.html


docs/styleguide/styles/_code.scss
```diff
@@ -18,7 +18,7 @@
             content: '';
             position: absolute;
             display: block;
-            z-index: 2000;
+            z-index: 1;
```
xak: @like-me1 this fixes the overlap!

PR #753, opened 2017-06-22
V2.0 gh pages deployment

derberg:
This is initial configuration. We are missing ssh key for travis push operation. I'll do it once you accept and merge this pull request.
- travis file is needed for travis to know what to do
- I noticed you use gulp extensively so there was no point to write custom sh script but just use available gulp plugins
- except of new gulp plugin I had to add bower to dependencies as only gulp was there

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: node_js
+node_js: 
+  - node
+install:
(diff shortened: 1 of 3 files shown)


xak:
:shipit:

Issue #771, opened 2017-06-28
Pkg: Build individual components


xak:
# Output
Standalone component CSS
- Build each SCSS file in `src/styles/components` into `dist/components`

joseegm:
Do we have a PR for this?

xak:
#773

joseegm:
### Review results

Component reviewed: https://github.com/SAP/techne/blob/7f5cb2db72aabf3a9d33c5d59e6a91b82f6a021b/dist/css/components/techne-card.css
and documentation at
http://sap.github.io/techne/www/components/cards.html - stand of 7, July 2017 12:00 UTC+1

Based on the [2nd Technical goal](https://github.com/SAP/techne/wiki/Techne-2.0-Goals): **Provide out-of-the box interoperability with Angular and React**.

1. **State classes cannot be isolated**.

    When managing an component **state** in Angular, is common practice to use the `ng-class`directive to map status variables to css classes, as follows:

  
    ```html
    <div class="tn-card" 
            ng-class="{ 'tn-card-readonly': !card.link, 'is-disabled': card.disabled }" />
    ```

    for this to work in the current implementation, it has to be rewritten as
    ```html
    <div 
            ng-class="{ 'tn-card': !card.link:, 'is-disabled': card.disabled , 
                        'tn-card-button': card.link }" />
    ```

   **NOTE: To match the current classes implementation extra work on the implementation side is needed.**

    **That extra condition on angular 1.5.x with 2 ways binding is going to be evaluated on every digest cycle.**

    **Decreasing the performance of the controller**


    Same problem with react, please see `className` [example](https://stackoverflow.com/questions/34521797/how-to-add-multiple-classes-to-a-reactjscomponent/34521945#34521945): 

    ```js
    var cardClasses = classNames({
          'tn-card': true,
          'tn-card--readonly': !self.link,
          'is-disabled': self.disabled
     });
    return (<div className={cardClasses} />);
    ```

     To mach the current implementation
     ```js
    var cardClasses = classNames({
          'tn-card': !self.link,
          'tn-card--button': self.link,
          'is-disabled': self.disabled
     });
    return (<div className={cardClasses} />);
    ```
    **NOTE: this will create an additional mutable property.**

    **Real live example from Builder public website**

     ```js
     <div class="cardRow" 
          ng-class="{'cardRowClickable' : clickable && !entry.entryClickDisabled && !entry.entryDisabled, 'cardRowDisabled' : entry.entryDisabled}" >
[...]

PR #773, opened 2017-06-28
#771 adds standalone components to pkg build prcoess

xak:


xak:
Running `gulp pkg-build --production` now also outputs standalone components

dist/css/components/techne-button.css
```diff
@@ -0,0 +1,66 @@
+/*! */
```
saadmhybris: 'button structure' comments are missing here. Should it be added to be consistent with other component css files?

xak: Yes, but not a show stopper. We haven't done buttons yet so it will be filled in later. Thanks!

PR #781, opened 2017-07-04
CCE-695 Correct help bubble

ghiobi:
Enforce use of techne help bubble.


saadmhybris:
Looks good!

saadmhybris:
@joseegm can we also account for this bug fix in the release notes?

joseegm:
@ghiobi 
There is something else there than readjusting the markup indentation?
Any changes to the class values etc?

PR #784, opened 2017-07-06
Feature/724

xak:
This add additional media types and actions to Cards. 

## NOTE: This PR removes the ability to use a single class. It become too unwieldy to account for every modification with the previous approach. The overhead isn't worth it.

We should discuss whether this is an admirable goal, i.e. to avoid needing `tn-card` and `tn-card--vertical`.

It would be easier to do this with an attribute selector but it would still require a specific way of coding ...

```
$block: ns(block);
//BLOCK
.#{$block} {
	//shared rules
	background: red;
	text-align: center;
	&--yellow {
		background: yellow;
	}
	&--green {
		background: green;
	}
}
//attribute
[class~="#{$block}--"] {
	@extend .#{$block};
}
//ELEMENTS
.#{$block}__element {
	font-size: 4em;
	&--flex {
		display: flex;
	}
	&--float {
		float: left;
	}
}
//attribute
[class~="#{$block}__element--"] {
	@extend .#{$block}__element;
}
```

This renders the below code where the class attribute selector picks up the modifier syntax and applies the base styles ...

```
.tn-block, [class~="tn-block--"] {
  background: red;
  text-align: center; }
  .tn-block--yellow {
    background: yellow; }
  .tn-block--green {
    background: green; }

.tn-block__element, [class~="tn-block__element--"] {
  display: block;
  font-size: 4em; }
  .tn-block__element--flex {
    display: flex; }
  .tn-block__element--float {
    float: left; }
```

It's a nice feature to have and I don't know of any libraries that do this. But the attribute selector is more expensive and I'd need to test the specificity.

Anyway, with this the Card will require multiple classes. We can always come back to optimizations later.

### DIFF
--- a/dist/css/components/techne-card.css
+++ b/dist/css/components/techne-card.css
@@ -5,14 +5,35 @@
         .tn-card__content
             .tn-card__header, .tn-card__description, .tn-card__label
 */
-.tn-card, .tn-card--button, .tn-card[role="button"] {
+.tn-card {
   font-size: 1rem;
   line-height: 1.375;
   color: #21262c;
   display: flex;
   max-width: 500px;
   border: solid 1px #d7d7d7; }
-  .tn-card__media, .tn-card__media--fill, .tn-card__media--round, .tn-card__media--hex {
+  .tn-card--vertical {
+    width: 370px;
+    flex-direction: column; }
+    .tn-card--vertical .tn-card__media {
+      width: 100%;
+      height: 370px; }
+  .tn-card--button, .tn-card[role="button"] {
+    cursor: default; }
+    .tn-card--button .tn-card__media, .tn-card[role="button"] .tn-card__media {
+      color: #009cdf; }
+  .tn-card--button, .tn-card[role="button"] {
+    cursor: pointer;
+    transition: border-color 0.2s, box-shadow 0.2s; }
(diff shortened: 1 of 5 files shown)


xak:
## NOTE: This PR removes the ability to use a single class. It become too unwieldy to account for every modification. The overhead isn't worth it.

We should discuss whether this is an admirable goal, i.e. to avoid needing `tn-card` and `tn-card--vertical.

It would be easier to do this with an attribute selector but it would still require a specific way of coding ...

```
$block: ns(block);
.#{$block} {
	//shared rules
	background: red;
    text-align: center;
	&--yellow {
		background: yellow;
	}
    &--green {
		background: green;
	}
}
//attribute
[class~="#{$block}--"] {
    @extend .#{$block};
}

.#{$block}__element {
    display: block;
    font-size: 4em;
    &--flex {
        display: flex;
    }
    &--float {
        float: left;
    }
}
//attribute
[class~="#{$block}__element--"] {
    @extend .#{$block}__element;
}

```

This renders ...

```
.tn-block, [class~="tn-block--"] {
  background: red;
  text-align: center; }
  .tn-block--yellow {
    background: yellow; }
  .tn-block--green {
    background: green; }

.tn-block__element, [class~="tn-block__element--"] {
  display: block;
  font-size: 4em; }
  .tn-block__element--flex {
    display: flex; }
  .tn-block__element--float {
    float: left; }
```

It's a nice feature to have and I don't know of any libraries that do this. But the attribute selector is more expensive and I'd need to test the specificity.


Anyway, with this the Card will require multiple classes. We can always come back to optimizations later.

saadmhybris:
all the modifiers are there, per design. looks good.

PR #788, opened 2017-07-10
Feature/728

xak:


like-me1:
@saadmhybris it came just recently not when the PR was created :) Once it ill be implemented it ill be easy to adjust, but I wouldn't mix those tasks together. Especially when this task is only for the website, not a package.

saadmhybris:
@xak @like-me1 -

never mind, i just scrolled down on the page and found the pallet table

saadmhybris:
@xak this is good to merge, but has conflicts now.

Issue #790, opened 2017-07-10
Pkg: Build modifications


xak:
Change package distribution ...

- CSS should build into the root of the `dist` directory
- `techne-all.css` should be named `techne.css`

PR #791, opened 2017-07-10
#790 moves css dist location to the root level; renames tecnhe-all.cs…

xak:
…s to techne.css


xak:
@saadmhybris This reorganizes the `dist` directory to put the CSS in the root. I also commented some of the tasks and CSS imports (e.g., icons) to only output the relevant CSS that is complete. 

You should expect to see ...

```
dist/
    components/
        card.css
        card.min.css
    techne.css
    techne.min.css
```

PR #792, opened 2017-07-10
Feature/779

xak:


xak:
@saadmhybris This goes with #790, the install script may be incorrect. Easy to change.

saadmhybris:
Looks good. I may change the version of 2.0.0-alpha.1 and mentions that it's a 2.0 release in the read me as well, but i'll take care of it when cutting the release branch.

Issue #796, opened 2017-07-11
Bug: Individual card.css is building in dev mode


xak:
No need for individual component CSS to build in dev mode. Only needed when building and releasing for production.

PR #797, opened 2017-07-11
#796 hotfix to prevent standalone components being build in dev mode

xak:


xak:
#796 — Prevents `card.css` from slipping in during regular development

Issue #798, opened 2017-07-11
Docs: Update Cards based on new color specs


xak:
Circle back to update the cards to reflect the new colors. Some colors are missing now.

joseegm:
@xak 
Where is the PR for this?

xak:
@joseegm If you look at the timeline above, the commits and the PR are all clearly reflected. It done, it's merged, it's fine, there's no problem here.

Saad found a bug during the code review and requested changes. The second commit was to fix that. Are you suggesting we merge the defect and then create a new issue and a new PR to fix the defect?

joseegm:
@xak 
No. I'm not suggesting that. What you described is exactly how it looks from the timeline. 

It is also a bit confusing since other code different from what's been done in this  issue has been committed to the same branch
https://github.com/SAP/techne/commit/8e56ccd4d061b5ecf762d1cf8353786d6cadf939#diff-763f975ef2d4cae80ac6e89925f32ed4L19
Since that's pretty prominent on the commit page and not noted on https://github.com/SAP/techne/pull/799#issuecomment-314595833
I though it was something else been merged after.

LeoT7508:
@xak 

The only issue I see with the cards is we need to add the interactions for the more "..." icon, and the "Generate Access Token" text button.

xak:
@LeoT7508 We need to circle back on those. The "More" is a contextual menu which hasn't been built and uses icons which are not done yet. I can add an actual button for "Generate Access Token" however. I'll put it on my list.

LeoT7508:
@xak 

We can talk more about the contextual menu and the Generate Access Token later. The rest of this looks spot on.

PR #799, opened 2017-07-11
#798 updates Cards colors based on new specs

xak:
See #798. Fixes regression bugs related to new color specs.

### DIFF
--- a/docs/styleguide/components/cards.html
+++ b/docs/styleguide/components/cards.html
@@ -15,7 +15,7 @@
 <div class="docs-section">
     <div class="docs-flow">
         <div class="docs-flow__box--12-5">
-            <h2 class="docs-header--h2">Structure</h2>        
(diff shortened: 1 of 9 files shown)


xak:
# NOT READY YET

xak:
# NOTES

- Changes some helper classes in docs HTML files
- Updates existing component SCSS to use the new `tn-color()` function and `@th-type` mixin
- Changes type helpers to use `tn-has-` prefix to be consistent with other helpers. The `is-` prefix should be reserved for interactive classes, not pure style helpers.
- Squeezed in a `dev-watch` fix where docs CSS changes were not browser syncing.

xak:
@saadmhybris The hover works for me locally. Note, the first card on the page should not have a hover. Can you double check?

screen shot 2017-07-12 at 12 37 45 pm

Issue #816, opened 2017-07-13
Pkg: Implement spacing system


xak:
Base unit 4px

- Include named spacing system with logical steps up from the base unit
- Create function
- Add helpers
- Add documentation on site

Related to #748

joseegm:
Looking at:

https://github.com/SAP/techne/commit/a8524e7b84c41ccb362fc7e9447058463512ddf6#diff-dc446cd5ee41bbcf85b1778fa17c006aR18

```scss
$tn-spacing: (
   base: $tn-spacing--base,    //4
   xs: $tn-spacing--base * 2,  //8
   s: $tn-spacing--base * 4,   //12
   reg: $tn-spacing--base * 6, //24 (should match bottom margin default)
   m: $tn-spacing--base * 8,   //32
   l: $tn-spacing--base * 10,  //40
   xl: $tn-spacing--base * 25, //100
   xxl: $tn-spacing--base * 37 //148
   ) !default;
```

I recommend **avoiding using literal identifiers for the spacing.** The spacing system is about proportions: a < 2a < 3a, 4a > 2a, etc.

The developers cannot calculate the proportions using a literal system, they would **have to learn then by memory, not by logic**.

1. The system used on the design is simpler and logical, illustrated bellow
    ![image](https://user-images.githubusercontent.com/22662903/28210348-2ff7c05c-6897-11e7-8209-cc6745285e8b.png)

    I will recommend using only one notation: *a, 2a, 3a, 4a, etc* for the spacing.

2.  Having 2 systems available, means the components will get developed, ones with one system other with the other, in the worst case a mix of both. **That is not consistent** and the developer have to remember what each system is.

3. When using the same notation as the design, we don't have to "map" one system to the other, they just correspond exactly: *2a = tn-space(2)*

**Only one spacing notation, less margin for error**.

**This can be simpler and effective**.

xak:
The function does support number arguments that multiply as you describe.  That, however, should be the secondary use case. The design system should adhere to a modular scale that will not always be simple increments of 4, esp. when you get to large numbers like 100. Therefore the "keywords" should be used more often than the numbers. 

[...]

PR #821, opened 2017-07-13
#816 adds spacing map, function, helpers, and docs

xak:
# NOTES
This adds the spacing specs to the library. 

The base unit is 4px and the Styles page includes convenience names for common spacing, like `reg` (24px) which will be `p` bottom margins and used a lot, `l` (40px) is common padding, `xl` (100px) is used for section bottom margins, etc. **THESE MAY CHANGE AS WE CONTINUE TO APPLY THE SYSTEM.**

## Function
The `tn-space()` function accepts these convenience keys — `base, xs, s, reg, m, l, xl, xxl` — and will return the appropriate value. If an invalid key is passed, it returns `reg` (24px).

The function also accepts a number — `tn-space(20)` — which will return the multiple of the base unit such as `4px * 20 = 80px`. The keys are preferred, but you can get any multiple of 4 if needed.

A second param can be passed to the function to indicate the value can be modified. If `true` is passed, this means that the value MAY be modified if the feature flag `$tn-space-modify` is set to `compact`. This allows an author to set, for example, the padding of a card using `padding: tn-space(l, true)` meaning this is `modifiable` but other spacing rules of the card may not be modifiable. Only those set to `true` would be affected **IF** that feature flag is enabled. Currently, the `compact` mode halves the `normal` values, i.e., 40px padding becomes 20px.

## Helpers
Classes are generated to override any margin or padding styles based on the keys. A `none` value is also included which can be handy if you need a quick way to remove margins and padding. 

The helpers take the form of `.has-padding-{key}, .has-margin-{key}, .has-padding-right-{key},  .has-padding-bottom-{key}`, etc. for `none, base, xs, s, reg, m, l, xl, xxl` **THESE MAY BE OVERKILL, WE CAN REVISIT LATER IF WE WANT TO REMOVE SOME OF THESE.**

## Docs
The system is documented on the Styles page.


xak:
I also updated the Cards SASS to use the new spacing function.

PR #831, opened 2017-07-18
CCE-696 add globe icon

SAPMontreal:


saadmhybris:
Regenerated and pushed the icon fonts files.

Issue #834, opened 2017-07-20
Docs: Handle nunjucks error


xak:
Never added error handler to HTML compilation

PR #835, opened 2017-07-20
#834 handles nunjucks errors

xak:
#834


xak:
Simple, simple fix to handle Nunjucks errors so Gulp doesn't crash. I just never added it before.

Issue #728, opened 2017-06-15
Docs: Implement color specs


xak:
Build out a page for the color palette. 

Update function based on new color specs.
https://sapse.invisionapp.com/d/main#/console/11153756/239338947/preview

Show usage of helper classes, examples ...
- `has-color-COLOR_NAME`
- `has-background-color-COLOR_NAME `

joseegm:
@xak 
Can you post the last version of the Colors here? Remember that the github repo is the main source of information. Invision is okay if you need it but is not the focus. Thx.

Don't forget to attach the PR for this.

joseegm:
@xak 
The Final PR is still missing here.

xak:
The PR was #788 and has been merged.

joseegm:
There is still a problem with the name "inverse" originally posted at: https://github.com/SAP/techne/issues/765#issuecomment-314693015

posting it again here for context:

**The name "inverse" doesn't reflect the usage. I'll include those colors inside of "Text" as well. Since they will be light colors, so they don't get confused by "Neutral"**

We need a more descriptive name.

PR #844, opened 2017-07-26
#728 chnages color inverse to text-inverse

xak:


src/styles/core/_settings.scss
```diff
@@ -104,7 +104,7 @@ $line-height: nth(map-get($type,0), 2) !default;
 
 //baseline
 $color: tn-color(text) !default;
-$background-color: tn-color(inverse) !default;
+$background-color: tn-color(text-inverse) !default;
```
joseegm: Using the text-inverse color group not for text. Should be using the neutrals color group.

saadmhybris:
@xak this branch also has conflicts now.

src/styles/components/card.scss
```diff
@@ -19,7 +19,7 @@ $block: ns(card);
     $tn-card-media-width: 75px !default;
     $tn-card-color-border: tn-color(neutral,2) !default;
     $tn-card-color-disabled: tn-color(text,3) !default;
-    $tn-card-background-color: tn-color(inverse) !default;
+    $tn-card-background-color: tn-color(text-inverse) !default;
```
xak: @joseegm This is more of a question for @LeoT7508. The `neutral-1` shade is not white. Do we need a backgrounds color palette or add 'white' to neutrals (which isn't perfectly logical)? 

Another option is to change `text` to something like `base` or `default` where it's more clear that `inverse` is meant to be the inverse of the base palette.

joseegm: Various points here:

1. I'm fine adding a new group of colors named `background`. That can include White. All themeable colors should be in color group anyway.
2. `text` is fine as it is, when is only used for text. `text` <> `text-inverse` is pretty clear what the colors in those groups are used for.

We cannot change names all the time. We need to stick with the scalability of the color system we selected. The system scale in two directions: more shades in a group, more colors groups.

xak: @joseegm The most important part of theming is that colors be identified by use not actual color. Beyond that, the design system is owned by design. The library architecture is designed to absorb changes but I agree that we need to narrow down the system as soon as possible.

I'm happy with system, I'll leave updates to you guys. Just please try to keep the palette limited. The larger it grows the harder it is to understand the use and theming becomes more difficult. 

Thanks.

joseegm: The color group names **are identified by use**. That's why using colors from the text-inverse group for background needs to be changed.

The color system as it is is scalable, no need to narrow it for now.

I'll talk with Leo and see how we gonna add the new color

xak:
Adjustments made based on recent discussion in #856.  Note that global [convenience variables](https://github.com/SAP/techne/pull/844/files#diff-dc446cd5ee41bbcf85b1778fa17c006aR122) are available for the status colors ... 

```
$tn-color--error: tn-color(status, 5) !default;
$tn-color--warning: tn-color(status, 3) !default;
$tn-color--success: tn-color(status, 1) !default;
```

joseegm:
@xak 
Putting the error, warning and success colors as global variables defeat the point of having named group colors.

1. Having variables and the `tn-color` function to get the colors: **are two ways to access the palette. It should be only one**.
2. When error have a second shade then you will need a second variable for it.
3. Accessing the colors via `tn-color(error, 1)` is consistent with the way we access all other colors.
4. **When introducing a new shade of error then all the other shades are going to be moved by +1**. This solution doesn't scale because then we have to change all the `$tn-color--warning: tn-color(status, 3)` to `$tn-color--warning: tn-color(status, 4)` instead of using only `tn-color(warning, 1)` and that doesn't needs to be updated if any other status get a new color.

@saadmhybris 
Please take a look at the aforementioned points, this solution doesn't scale and we agreed on the last meeting we need to separate the different statuses in individual groups. 

Don't see why we need to move away from that agreement.

xak:
@joseegm With all due respect this has nothing to do with any of your roles on this project. The design system is defined by the design team and implemented by the development team. How does this affect you in any way? You must see that you are dictating implementation details which POs should not do.

You need to provide some leeway for the dev team to work out the design system implementation. The system will evolve and change, please give us some room to test out ideas. I'll take feedback from the developers using the toolkit ... if something is confusing or not working well, we will adjust it. There is no problem here, you are creating rules based on your narrow understanding of front-end architecture and ignoring the expertise of others. This is not your decision to make.

You can document the colors however you please on the website, If you want to break out the statuses into three groups, fine. But the SASS implementation can differ if it is more logical in the toolkit. Applications often use status codes, i.e., 0 is success, 1 is error, so consider our SASS application the same way, i.e., 1 is success, 3 is warn, 5 is error. 

I think we can agree that we need to nail down the colors. My solution to your scaling problem ... **DON'T add more colors.** I mean, how many shades of error status do we really need? The color palette should now have enough colors to realize every component. The challenge now for design is to apply the color palette to new components, not expand it for new components. It will change, I understand that and we will handle that then. "Find and replace" is a powerful tool ... I'm quite good at it.

Are you going to continue to hold this up or let us proceed?

joseegm:
@xak 
I'm noticing the pitfalls on the implementation. As PO I have to make sure that the Systems we already have in place are used correctly. In this case they are not used correctly and the implementation solution is not scalable.

It does affect the project if after a while we need to split the status colors in color groups anyway. Better to do it now.

No need to create new systems like this new global status color variables. **That increase the complexity.**

As on my previous comment, **I suggested to @saadmhybris that take in count the points I mentioned. Is his decision if this solution is going to be used.** He is accountable if the decision taken now, implies more overhead and changes down-the-line.

Your solution of "not adding more colors" is completely impossible and not focused on reality. The colors are added on needed basis, we need more warnings and error colors to satisfy the design of the error messages. The Color Systems is escalable in two dimentions, that's not the problem.

xak:
You are "noticing the pitfalls on the implementation" based on your narrow viewpoint. You're trying to solve problems that we don't have. 

Explain to me how it makes sense to elevate the status colors individually to the same level as `text`, `backgrounds`, and `actions`? The concept of "grouping by usage" breaks completely. It shows a weakness in the classification and is simply not logical. You want to treat `success`, `warning`, and `error` as first-class citizens of the color palette for what reason? Only so you can have a second error shade that gets used maybe once in the whole library. How do you rationalize these as top-level groups.

Every color decision must be considered in terms of theming. If colors are "added on needed basis, we need more warnings and error colors to satisfy the design of the error messages" then where does it end? When we have 63 colors and 74 background colors like we have now? How do we map colors to a new theme when a background color is used once just because it looks better to the designer? How do you implement a theme when a tree uses one color for borders and a table uses another just because the designer thought 90% lighter looks a little bit better? This isn't how a design system should work.

Look at any Apple-made iOS application like Mail or Notes. How may colors do they use? 6 or 9 maybe, max. A controlled color system is constraining but it can be very powerful when done correctly. Your designer-has-total-freedom-to-chose approach is what will cause problems down the line, not global variables.

You're holding this up for no good reason. As I said, document it however you want on the site but there is nothing wrong with the SASS implementation. And if problems show up after devs start using the toolkit then **I** will fix them. But I've taken these things into account. I have many years of experience adapting the front-end to changing whims of designers and I've learned one or two things that help absorb the shock of changing minds mid-stream. Please start showing some trust in your fellow team members.

joseegm:
@xak 
We already have the same different groups names for "Text" and "Text-Inverse" for exactly the same reason. **That's why separating the different status makes total sense.**

Exactly because of theming we should have **every color in the palette and not use "calculated" shades of the same colors that may not works out of the box for a different theme.** Theming is a core feature of the Color System, because it gives the flexibility to change any color independently.

**You are comparing apples to oranges again: The design language of iOS is not the same as the one we are building. **

The statement: **"And if problems show up after devs start using the toolkit then I will fix them."** is exactly the type of mentality and approach  I try to avoid. That simply is: **creating technical dept because the solution wasn't architected correctly.

As said before. I make my points clear on https://github.com/SAP/techne/pull/844#issuecomment-320201319 - that BTW hasn't been answered clearly.

**I'm not holding anything** the points were made, to make @saadmhybris aware of the pitfalls. I strongly think the current implementation is not scalable and unnecessary; and I'm trying to avoid changing it more down the line.

LeoT7508:
Saad is leading development on the project with guidance from Zack. From a design side this works. Per the team discussion and the new methodology suggested by Irla the final call rests in the development teams hands. Let's move forward with their recommendation and close this issue.

saadmhybris:
Let's all revisit the RACI chart and stick to our responsibilities. We need to keep moving, even if the solutions are not perfect, let's put it out there for teams work with. We can come back and fix it if needed.

joseegm:
@saadmhybris 
I'm fine with the decision, just wanted to make sure all aspects are taken in count and we are not building unnecessary tech debt.

Issue #714, opened 2017-06-12
Pkg: Namespace Techne variables


xak:
Variables should be namespaced.

`$tn-fonts`, `$tn-color--link`, etc.

PR #845, opened 2017-07-26
#714 namespaces all variables

xak:
#714


xak:
This simply goes through and adds the `tn-` prefix to all variables that did not have it.

Issue #687, opened 2017-05-23
Implement a component-only spec sheet 


xak:
An HTML page with all components would be good for testing purposes. This spec sheet should be devoid of any additional styles and ONLY include `components.css`. This will allow us to see if any additional stylesheets are interfering with the component styles.

joseegm:
This will also help with automated visual testing, mainly we only need to test against this page.

PR #860, opened 2017-08-08
Feature/687

xak:
#687


xak:
This is the first pass at a standalone component development process. Looks the `README` in `test` to start up the server. There are custom Nunjucks filters in `app.js` to compile the SASS and handle modifier classes. More of these can be added as needed.

The cards are built with a Nunjucks macro, but the idea is that static HTML could be used if that speeds up development. The only requirement is that the component page should reflect all the component types represented by the design deliverables.

Any developer should be able to add a new page, create a new SCSS file and start developing. Once complete, the HTML can be copied and used on in the `docs`. This opens up a new stream of work separating the lib from the docs.

**Note: the `card/data.json` file is not currently used but included to illustrate the optimal structure.** We should work to define a standard structure that will be used for all components. My approach separates the content data into `properties`, presentation types are part of a `modifier` object which has `block` and `ELEMENT` arrays, and then separates `state` and `aria` into their own objects.

```
{
    "properties": {
        "title": "CARD_HEADER",
        "description": "CARD_DESCRIPTION",
        "status": "CARD_STATUS",
        "media": {
            "image": "https://techne.yaas.io/images/product-thumbnail-wide.png",
            "alt": "CARD_MEDIA_ALT"
        },
        "actions": "",
        "status": "",
        "body": ""
    },
    "modifier": {
        "block": ["button", "vertical"],
        "media": ["round",  "fill"]
    },
    "state": {
        "disabled": false
    },
    "aria": {
        "role": "button"
    }
}
```

The more we can rally around a consistent vocabulary the better, i.e., using `title`, `description`, etc.

saadmhybris:
This is super cool!!

Issue #875, opened 2017-08-14
Pkg: Implement input groups


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `dev`

> This story is being separated from the `form` story since they are distinct

# Inputs
**Design**
Background at https://github.com/SAP/techne/wiki/Form-Inputs
https://app.box.com/folder/14968986610

**Specs**
- **Invision:** https://sapse.invisionapp.com/d/main#/console/11153756/240329805/preview
- **Zeplin:** https://zpl.io/a3ozkNa

### Input group
https://github.com/SAP/techne/wiki/Component-Specs:-Inputs#inputs-1
```
.tn-input-group
	.tn-input-group__addon?+(--step-up,--step-down)
```

# Outputs

- Account for text, icon and button addons on left and right of inputs
- Account for spinner buttons on right of `input[type=number]` inputs

Issue #795, opened 2017-07-11
Pkg: Implement forms


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge

# Inputs

**Design**
Background at https://github.com/SAP/techne/wiki/Form-Inputs
https://app.box.com/folder/14968986610

**Specs**
https://sapse.invisionapp.com/d/main#/console/11153756/240329805/preview
https://sapse.invisionapp.com/d/main#/console/11153756/240329804/preview

> This story relies on some `form-group` structures not fully defined yet. For example, all form fields will be styled based on the `form__control` class, not individual classes for each type, i.e., `select`, `input`, `textarea`

### Form
https://github.com/SAP/techne/wiki/Component-Specs:-Inputs#forms-layout
```
.tn-form
	.tn-form__group
		.tn-form__control+((.is-disabled|[disabled]) | :focus | .is-valid | .is-error | (.is-readonly|[readonly]))
```

### <s>Input group</s>
https://github.com/SAP/techne/wiki/Component-Specs:-Inputs#inputs-1
```
.tn-input-group
	.tn-input-group__addon?+(--btn-up,--btn-down)
```

# Outputs

- 100% width inputs
- <s>Account for text, icon and button addons on left and right of inputs</s>
- <s>Account for spinner buttons on right of `input[type=number]` inputs</s>
- Style states of inputs — disabled, readonly, focus, valid, error
- Forms page on website

xak:
The tooltip is being deferred until the tooltip component has been designed.

LeoT7508:
@xak 

2 pieces of feedback. 

For disabled fields - It works fine I just noticed that for buttons you use a crossed out icon to show you can't click on the field. Should we do that here to? I'd suggest going on way or the other. 

For check boxes and radio buttons. The circle option and the check box should both be 26px squared. I think right not you have the radio buttons as 32px. The check boxes look good.

xak:
@LeoT7508 

Fixed ...
https://github.com/SAP/techne/pull/876/files#diff-9b2d847fc5562e255de1ce5050bf5c6bR65
https://github.com/SAP/techne/pull/876/files#diff-9b2d847fc5562e255de1ce5050bf5c6bR9

LeoT7508:
@xak 

Thanks, we should be good to go here.

PR #876, opened 2017-08-14
#795 styles for inputs and textarea; adds form component

xak:
#795 

**All examples can be seen at [localhost:3030/form](http://localhost:3030/form) after starting with `npm test`**

This story implements the `core/forms` styles for all `input` types, including `radio` and `checkbox` and `textarea`.

It also implements the `form` component which primarily arranges the elements inside the `form__group` to stack the label, show the input, put the message below, etc. For this particular component,  the `form` block can be implied, meaning it is not required. In most instances, the `form__group` can be considered the root.

Compare to https://zpl.io/2Gvzk9b, https://zpl.io/25vykjV and https://zpl.io/boeBWWV

This is great but not required ...
```
<form class="tn-form">
  <div class="tn-form__group"></div>
  <div class="tn-form__group"></div>
</form>
```

This is just as good ...
```
  <div class="tn-form__group"></div>
  <div class="tn-form__group"></div>
```

Modifiers may be added to the `form` block later on, currently there are none in place.

## NOTES
- The various types of input text fields have been separated to #875 to cut down on the complexity
- Tooltips next to the label will be added after the tooltip component has been designed
- `input[type=search]` and `input[type=number]` need additional work and will be addressed with the `input-group` story
- There are inconsistencies in how `required` fields are marked in the designs between `radio` and `checkbox`. Marking an individual `label` will bold it, same goes for the `legend`, an `*` must be added manually.
- The default message — not shown — includes no icon, modifiers `help`, `error` or `warning` add the icon to the message.

### DIFF
--- a/ops/misc/src-tmpl.html
+++ b/ops/misc/src-tmpl.html
@@ -1,19 +1,17 @@
 <!--
 __COMPONENT_METHOD__:
-    label="", (string)
-    description="", (string | optional)
-    modifier="" (string | optional)
+    properties={},
+    modifier={},
+    state={},
+    aria={}
 -->
-{% macro __COMPONENT_METHOD__(
-    label="",
-    description="",
-    modifier="") %}
-<div class="tn-__COMPONENT_ID__{{ '--' + modifier if modifier }}">
+{% macro __COMPONENT_METHOD__(properties={}, modifier={}, state={}, aria={}) %}
+<div class="tn-__COMPONENT_ID__{{ modifier.block | modifier('__COMPONENT_ID__') }}">
     <p>
-        {{ label }}
-        {% if description %}
+        {{ properties.label }}
+        {%- if properties.description %}
             <br>
-            {{ description }}
+            {{ properties.description }}
         {% endif %}
     </p>
 </div>
--- a/ops/misc/src-tmpl.json
+++ b/ops/misc/src-tmpl.json
@@ -1,12 +1,18 @@
 {
     "id": "__COMPONENT_ID__",
     "name": "__COMPONENT_NAME__",
-    "modifiers": [
-        "no-border"
-    ],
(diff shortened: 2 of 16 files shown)


xak:
@saadmhybris You can wait to merge this after Leo reviews the design or merge to a review branch as we discussed.

saadmhybris:
@xak ok, I'll just wait for Leo.

There one small issue I found, but most likely templated related - 

screen shot 2017-08-15 at 9 26 16 am

not sure if the error message is supposed to stretch out beyond the first name input?

xak:
@saadmhybris The form fields and the message containers are blocks but do not have built-in widths and will always fill the parent container. In that example the fields float so they are not 100%. I set it up so that a message can be used with the `form__group` or the `form__item` so in this example I wanted to show how it works with the group, primarily that it's a little strange to be flush with the inputs.

I assume that in the case of a city, state, zip if all are empty we may want to show a single message under them whereas if the city is empty it may show only under the city. However, in my experience, having been down this road before, trying to show an error message under a two-character state field is problematic so we'd likely recommend single message for all three fields in that set.

I made it as flexible as possible to account for a variety of use cases in the absence of specific UX specifications.

saadmhybris:
@xak - totally agree that the error message should take the width of the parent container. also agree with the use case of showing a group error message. 

Maybe we need to consider an alternate approach to field group error message. as it is built right now, it looks broken from a visual stand point and the implementor would have to do some sort of CSS overwrite to fix it. 

from a code prespective, looks good.

xak:
@LeoT7508 This is ready for your review. See https://github.com/SAP/techne/wiki/Development-Guidelines#design-reviewer

LeoT7508:
@xak @saadmhybris 

This checks out from a design side.

PR #882, opened 2017-08-17
#875 adds input-group component with addon options; adds button mixin

xak:
#875

> This is currently compared with `feature/795` since it builds on that. I will rebase once that branch is merged.

# Review Notes
### As discussed previously it is not possible with CSS alone to 1) maintain all states of the input field, i.e., hover, active, and 2) have the border wrap the "add-on" content. 

I implemented as Bootstrap does it with standalone-ish add-ons.

screen shot 2017-08-17 at 3 39 42 pm

The design could be implemented if the widths of the `addon` containers are fixed regardless of the content — e.g. "$" and "km/h". The input field content could be padded the add-on content would be shifted on top. However, this would require addt'l classes on the parent container to "tell" the children how they need to behave, which is not desirable. As it is right now, you only have to include the add-on containers if you want them ...
```
<div class="tn-input-group"> 
    <span class="tn-input-group__addon">$</span>
    <input value="">
    <span class="tn-input-group__addon">km/h</span>
</div>
```
vs.
```
<div class="tn-input-group tn-input-group--with-before tn-input-group--with-after"> 
    <span class="tn-input-group__addon">$</span>
    <input value="">
    <span class="tn-input-group__addon">km/h</span>
</div>
```

### Icon only buttons are not working properly, should include these as button types

With the existing buttons it is possible to use only an icon and not include a label but the styling is not correct with an extra margin appearing on the right of the icon.
 
screen shot 2017-08-17 at 3 41 31 pm

I recommend adding a new "icon only" button type which would be used here — and would be useful in other places too. There shouldn't be any special CSS needed here for the buttons, it should be a button that could stand on it's own so I'm leaving the margin for now and we can look at the new button separately.

### DIFF
--- a/src/styles/components.scss
+++ b/src/styles/components.scss
@@ -1,3 +1,4 @@
 @import "components/button";
 @import "components/card";
 @import "components/form";
+@import "components/input-group";
--- a/src/styles/components/_mixins.scss
+++ b/src/styles/components/_mixins.scss
@@ -1,0 +1,17 @@
+$ns: ns();
+@mixin tn-button-reset {
+    margin: 0;
+    padding: 0;
+    font-smoothing: antialiased;
+    appearance: none;
+    outline: 0;
+    border: 0;
+    display: inline-block;
+    text-decoration: none;
+    cursor: pointer;
+    user-select: none;
+    vertical-align: middle;
+    white-space: nowrap;
+    background-color: transparent;
+
+}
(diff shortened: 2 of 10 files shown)


xak:
These are ready for review ... this is branch `feature/875`

All flavors in the design spec are represented — https://zpl.io/a3ozkNa

xak:
@LeoT7508 I just pushed a change to include the background on the add-on container.

LeoT7508:
@xak 

Just reviewed the add-on container, looks perfect.

PR #909, opened 2017-08-24
Feature/872 - Pkg: Icons

saadmhybris:
icons can be viewed with `gulp test `


saadmhybris:
Closing this PR as we need to change the icon system back to font icons as SVG system does not allow for an easy way to change color.

Issue #897, opened 2017-08-23
Pkg: Implement new button types


xak:
- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge

# New

* Large button with icon to the left (standard and text versions)
* Icon only buttons in all 3 sizes (with label and non-label option for large)

# Changes

* Large buttons should be 76px height
* Make text buttons same height as standard
* Make `is-active` a class state

PR #910, opened 2017-08-24
Feature/897

xak:
#897

### DIFF
--- a/ops/misc/test-component.njk
+++ b/ops/misc/test-component.njk
@@ -6,13 +6,12 @@ __COMPONENT_METHOD__:
     aria={}
 -->
 {% macro __COMPONENT_METHOD__(properties={}, modifier={}, state={}, aria={}) -%}
-<div class="tn-__COMPONENT_ID__{{ modifier.block | modifier('__COMPONENT_ID__') }}{{ ' is-disabled' if state.disabled }}"{{ ' aria-disabled="'+ aria.disabled +'"' if aria.disabled }}>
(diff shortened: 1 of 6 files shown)


xak:
@LeoT7508 The new button types are done. You can review in branch `feature/897` at http://localhost:3030/button

**A few notes**
- I added a new `--link` button with the shorter heights and the underlines.
- The `--text` button is now the same height as the regular button and does not have have underlines —  it's not easy to do the underlines and allow icons with a single class. _We should reconsider the name "Text Button" since the difference is more about the background color._ It becomes sort of silly when you have an `--icon`-only `--text` button. Maybe "secondary" or "clear" or something like that.

```
<button class="tn-button tn-button--text tn-button--icon" aria-label="BUTTON_LABEL">
    <span class="tn-button__icon tn-icon tn-icon--foo" role="presentation">✕</span>
</button>
```
- I created an `--action-bar` button for that special case. We can consider a different name if we find this button is applicable in other instances. Maybe `--icon-with-label` would work but I'm fine with action bar for now.

LeoT7508:
@xak 

The only one I'm not sure about is the different between a link button and a text button. The hovers, active, idle, disable states all work right. Colors are right, sizing is right. 

Looks good from a design side.

xak:
@LeoT7508 "Link" vs. "Text" is part of why I think we need to rename the "Text" button.

"Text" is just a different background (transparent) and text color ... whereas the "Link" button is shorter and works more like, well, a link ... it will play nicely with paragraph text as you noted before.

LeoT7508:
@xak 

We can call it whatever you think is best. As long as the styling is the same we should be good.

xak:
I'll just leave it as is for now, we can assess later if we need more buttons.

Issue #898, opened 2017-08-23
Pkg: Sync type specs


xak:
Some type changes have occurred, need to double check the base CSS settings.

Update in `docs` as well

PR #911, opened 2017-08-24
#898 updates line-height to match upated specs

xak:
#898 

Line heights that did not match the base unit of 4px were updated — like 22px is now 24px, etc.


saadmhybris:
looks good.

Issue #906, opened 2017-08-24
Pkg: Add card groupings


xak:
- [x] Develop
- [x] Design review
- [ ] Code review
- [ ] Merge

The cards cannot have a fixed width so either need to match the implementation in 1.5 or build out `card-group` component.

Initial implementation should cover the 3 and 2 col layouts. See Zeplin specs at https://zpl.io/aBxD7Q2

PR #912, opened 2017-08-25
#906 adds card-group component with some adjustments to card styles

xak:
#906 

Find in branch `feature/906` at http://localhost:3030/card-group. This implements default 3-per row and an option for 2-per row. 

<s>Still need to handle the Product Cards which are 4-per row but these can go through now.</s>

**I went ahead and added a 4-col option** Basic responsive behavior is included to stack cards under `s` size — 800px.


LeoT7508:
@xak 

Don't we need to show a card with padding around the image?

We have an example of that in the visuals. 

Other than that these look good.

xak:
The circle image is on the Cards page .. this page is meant to show the layout so the Cards inside can vary. We can show whatever combos we want on the website.

xak:
@saadmhybris Responsive query has been added to stack under 800px.

Issue #904, opened 2017-08-24
Docs: Update buttons based on new types


xak:
Blocked by #897

PR #919, opened 2017-08-25
#904: updated buttons based on design updates.

saadmhybris:


xak:
:shipit: 

We'll have to update again with the icons but looks good for now.

PR #925, opened 2017-08-28
Feature/921

xak:
See `icons/README` The icon font has been renamed to prevent conflicts with the 1.5 version. I also removed the `src` directory and did some associated cleanup. I can address the conflicts once reviewed.

# NOTE: I did not update the docs tasks. 

The icons currently work with the test framework and running `gulp build:dist --production` will copy everything to `dist` (dev mode copies to `tmp`) with the CSS pathed to assume the same level directory. 

The docs tasks will need to copy the fonts to the appropriate directory.


saadmhybris:
@xak 

- branch has conflicts since we have a few PR merges 
- test index file missing link to icons http://localhost:3030/

xak:
@saadmhybris fixed!

https://github.com/SAP/techne/pull/925/files#diff-24ff85c62613841bf17b9d8f431e2febR17

Issue #761, opened 2017-06-23
Pkg:  Distributed CSS files should have header with library version and copyright


xak:
A comment at the top of each distributed file.
```
/*!
 * Techne v{VERSION_NUMBER}
 * Copyright (c) {RELEASE_YEAR} SAP SE or an SAP affiliate company.
 * Licensed under Apache License 2.0 (https://github.com/SAP/techne/blob/master/LICENSE)
 */
```

See https://github.com/tracker1/gulp-header

PR #930, opened 2017-08-28
Feature/761

xak:
#761


dist/techne.css
```diff
@@ -1,9 +1,9 @@
...
 /*!
- * Techne v2.0
- * Copyright (c) 2017 SAP SE or an SAP affiliate company.
- * Licensed under Apache License 2.0 (https://github.com/SAP/techne/blob/master/LICENSE)
- */
+* Techne v2.0.0
+* Copyright (c) 2017 SAP SE or an SAP affiliate company.
+* Licensed under Apache License 2.0 (https://github.com/SAP/techne/blob/master/LICENSE)
+*/
+@charset "UTF-8";
```
xak: **Note:** Technically the `@charset` needs to be the first thing in the file but you cannot disable this in `gulp-sass`. It is automatically added because of the icon encodings.

A different solution needs to be devised if we want to be 100% fully compliant. To my knowledge, however, this `@charset` declaration is not heavily relied upon by user agents so the impact is probably minimal with this current solution.

dist/components.css
```diff
@@ -1,4 +1,9 @@
 /*!
+* Techne v2.0.0
```
saadmhybris: should the version number be v2.0.0-alpha.1 to match what we'll have in the package.json?

Issue #914, opened 2017-08-25
Pkg: Implement status indicators


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merged to `v2.0`

# Inputs
- See background at #814
- See [1.5 implementation](https://techne.yaas.io/components/Status-Indicators.html)

## Specs
- **Zeplin:** https://zpl.io/VQ3vwkb
- **PSD**: https://app.box.com/folder/14968986610

## Basic block structure (suggested)

Implement labels and badges
```
.tn-label+((--success |--warning | --error))
.tn-badge+((--success |--warning | --error) | --pill)
```

# Outputs

- Match 1.5 functionality
- Since this is an [enhancement of a Bootstrap component](https://getbootstrap.com/docs/4.0/components/badge/), additional Bootstrap modifications may need to be accounted for. See [v4](https://getbootstrap.com/docs/4.0/components/badge/)

PR #937, opened 2017-08-29
#914 labels and badges

xak:
#914

See branch `feature/914`. Two pages http://localhost:3030/label and http://localhost:3030/badge


LeoT7508:
@xak 

All good from my end. Thanks.

Issue #885, opened 2017-08-21
Pkg: Implement alert component


xak:
# Workflow

- [x] Develop
- [ ] Design review
- [x] Code review
- [ ] Merge to `dev`

# Inputs
- See background at #813
- See [1.5 implementation](https://techne.yaas.io/components/Alerts-Errors-Notifications.html)

## Specs
- **PSD**: https://app.box.com/folder/14968986610
- **Zeplin:** https://zpl.io/VxA6xka

## Basic block structure (suggested)

```
.tn-alert
	.tn-alert__action
```

# Outputs

- Match 1.5 functionality
- Since this is an [enhancement of a Bootstrap component](https://getbootstrap.com/docs/3.3/components/#alerts), additional Bootstrap modifications may need to be accounted for. See [v4](https://getbootstrap.com/docs/4.0/components/alerts/)

Issue #855, opened 2017-08-01
Docs: Migrate Techne Website delivery to Jekyll


saadmhybris:
Goal: Use Jekyll to power techne documentation site.

**Current Status:** 
- [V2.0 branch](https://github.com/SAP/techne/tree/v2.0) has the most up to date Techne 2.0 component code
- There is already separation between the library and the documentation site 

**Tasks for completing the migration to Jekyll:** 
- [x] Replace ./docs folder with Jekyll
- [x] Integrate Jekyll server with Gulp to serve local dev site 
- [x] Setup Barebone structure for the site. 
- [x] migrate content from the previous version of the site (old v2.0 content that was already done) 
- [x] Compile SCSS code with gulp for both the library and documentation site. 
- [x] Live reload for local development 
- [x] Code snippet component should be driven by the component code to reduce redundancy. 
- [x] Port 1.5 components that don't require redesign - added as a separate task this is a relatively big undertaking. discussions to continue here -  #858
- [x] Deployment for dev, staging, prod - moving this out to its own task as this should not be a blocker for this ticket to be completed and merged #857
 
Development branch for this task is not under feature/855-jekyll

saadmhybris:
FYI Jekyll will not regenerate when .yml files are touched. 

Issue discussed here - https://github.com/jekyll/jekyll/issues/3854

saadmhybris:
Changed the goal of the task to just the migration to Jekyll to serve up the documentation site. 

1.5 component migration and MVP discussion to be handled in separate tickets.

PR #962, opened 2017-08-31
#855: Pkg - Alerts

saadmhybris:
#885

### DIFF
--- a/scss/components/alert.scss
+++ b/scss/components/alert.scss
@@ -5,7 +5,7 @@
 
 /*!
 .tn-alert+(--warning. --error)
-    .tn-alert__action
+    .tn-alert__close
 */
 $block: ns(alert);
 
@@ -14,6 +14,7 @@ $block: ns(alert);
     $tn-alert-background-color: tn-color(text, 3) !default;
     $tn-alert-background-color--warning: tn-color(status, 2) !default;
     $tn-alert-background-color--error: tn-color(status, 3) !default;
+    $tn-alert-color--warning: tn-color(text, 1) !default;
     $tn-alert-color: tn-color(text-inverse, 1) !default;
     $tn-alert-border-color: tn-color(neutral, 2) !default;
     $tn-alert-height: tn-space(13) !default;
@@ -25,37 +26,22 @@ $block: ns(alert);
     background-color: $tn-alert-background-color;
     min-height: $tn-alert-height;
     padding: tn-space(s);
-    // padding-right: $tn-alert-action--clerance;
 
-    // Element
-    // &__action {
-    //     float: right;
-    //     margin: -(tn-space(s));
-    //     margin-right: -$tn-alert-action--clerance;
-    //     .tn-button{
-    //       color: tn-color(text-inverse, 1);
-    //       min-height: $tn-alert-height;
-    //       &:hover,
-    //       &:active,
-    //       &:focus{
-    //         color: tn-color(text-inverse, 1);
-    //       }
-    //     }
-    // }
-    &__action {
-        float: right;
-        margin: - (tn-space(s)/2);
-    }
+    // Elements
     &__close {
         @include tn-button-reset;  //removes default button chrome
         color: tn-color(text-inverse, 1);
         width: tn-space(m); //best practice hit space
         height: tn-space(m);
-}
+        float: right;
+        margin: - (tn-space(s)/2);
+    }
 
     // Modifiers
     &--warning {
       background-color: $tn-alert-background-color--warning;
+      color: $tn-alert-color--warning;
+
     }
     &--error {
       background-color: $tn-alert-background-color--error;
--- a/scss/components/alert.scss
+++ b/scss/components/alert.scss
@@ -1,0 +1,52 @@
+@import "../core/settings";
+@import "../core/mixins";
+@import "../core/functions";
+
+/*!
+.tn-alert+(--notification. --error)
+    .tn-alert__action
+*/
+$block: ns(alert);
+
+.#{$block} {
+    //SETTINGS
+    $tn-alert-background-color: tn-color(text, 3) !default;
+    $tn-alert-background-color--notification: tn-color(status, 2) !default;
+    $tn-alert-background-color--error: tn-color(status, 3) !default;
+    $tn-alert-color: tn-color(text-inverse, 1) !default;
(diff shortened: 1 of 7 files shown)


scss/components/alert.scss
```diff
@@ -0,0 +1,52 @@
...
+    background-color: $tn-alert-background-color;
+    height: $tn-alert-height;
+    padding: tn-space(s);
+    text-align: center;
+
+    // Element
+    &__action {
+        float: right;
+        margin: -(tn-space(s));
+        .tn-button{
```
xak: We want to avoid this ... changing components based on their parent. 

So there are a few options ...

1. Add a modification for buttons so you can get the inverse type without a background - `.tn-button--clear` or `tn-button--inverse`
2.  Add a button mixin where you can embed button styles on the fly into different elements - `.z-button { @include tn-button($color, $background) }`
3. Create a specific button for this component.

Since we only have a single use case thus far for an inverse button I don't recommend #1, same for #2 although we may need this later if we need different buttons with different backgrounds, so #3 is the best based on the info we have at this time.

I recommend creating a `.tn-alert__close` button that has an icon inside. Then the CSS here could be ..

```
    &__action {
        float: right;
        margin: tn-space(2) -(tn-space(s));
    }
    &__close {
        @include tn-button-reset;  //removes default button chrome
        color: tn-color(text-inverse, 1);
        width: tn-space(m); //best practice hit space
        height: tn-space(m);
}
```

The button-reset mixin is part of component mixins. See above for `@import`

scss/components/alert.scss
```diff
@@ -0,0 +1,52 @@
+@import "../core/settings";
+@import "../core/mixins";
+@import "../core/functions";
+
```
xak: Add `@import "../components/mixins";` to get the `tn-button-reset()` mixin.

scss/components/alert.scss
```diff
@@ -0,0 +1,52 @@
...
+    $tn-alert-background-color--error: tn-color(status, 3) !default;
+    $tn-alert-color: tn-color(text-inverse, 1) !default;
+    $tn-alert-border-color: tn-color(neutral, 2) !default;
+    $tn-alert-height: tn-space(13) !default;
+
+    // Block
+    color: $tn-alert-color;
+    border: solid 1px $tn-alert-border-color;
+    background-color: $tn-alert-background-color;
+    height: $tn-alert-height;
```
xak: Should be min-height ... always test variable content length.

screen shot 2017-08-31 at 2 38 06 pm

scss/components/alert.scss
```diff
@@ -0,0 +1,52 @@
...
+/*!
+.tn-alert+(--notification. --error)
+    .tn-alert__action
+*/
+$block: ns(alert);
+
+.#{$block} {
+    //SETTINGS
+    $tn-alert-background-color: tn-color(text, 3) !default;
+    $tn-alert-background-color--notification: tn-color(status, 2) !default;
```
xak: I use `--warning` in every other instance, let's be consistent.

test/templates/alert/component.njk
```diff
@@ -0,0 +1,16 @@
+{#
+    see ./data.json for object structure
+#}
+
+{% from "../button/component.njk" import button %}
+{% macro alert(properties={}, modifier={}) -%}
+<div class="tn-alert{{ modifier.block | modifier('alert') }}" role="alert">
+  <div class="tn-alert__action">
+    {{ button({ icon: 'remove' }, modifier={ block: ['icon', 'text', 'small'] }, aria={ label: 'Remove' }) }}
```
xak: To simplify this and avoid the button styling issue, you could add a custom button element with an icon ...

```
<button class="tn-alert__close" aria-controls="{{properties.id}}">{{ icon('remove') }}</button>
```

You'll have to import icon instead of button `{% from "../icon/component.njk" import icon %}`. 

And consider putting an `id="{{properties.id}}"` on the alert with the close button using `aria-controls="{{properties.id}}"`.

> The id could also be randomly generated. It's important that they match but have no significance to a developer so passing it explicitly simply adds overhead.

test/templates/alert/data.json
```diff
@@ -0,0 +1,9 @@
+{
+    "properties": {
+        "text": "Alert Text",
+        "action": "#/action"
```
xak: You're not explicitly passing an action here so probably a boolean like "dismissible" would work — some flag to either include the close or not. See https://getbootstrap.com/docs/4.0/components/alerts/#dismissing. 

But I believe currently ALL of our alerts can be dismissed so for MVP it's OK.

LeoT7508:
@xak 

We can right align sure. I don't think those messages should be paragraphs either way. Might be important restrict the amount of type you allow in that alert.

saadmhybris:
@LeoT7508 I strongly feel that white on orange works better here than black on orange. Should we stick with white? 

screen shot 2017-08-31 at 3 44 39 pm

xak:
@saadmhybris One change re: the logic for the close button. 

For future reference, these may be helpful ...

To control the white space ... which unfortunately has to be done manually.
http://mozilla.github.io/nunjucks/templating.html#whitespace-control

There is also an indent filter to will indent all lines ... 4 spaces is the default
```
{{ button({ label: "Click me"}) | indent(8) }}
```

xak:
Also, thinking about what we discussed yesterday and I see that injecting additional classes into a component — like with your `action` and `button` — will be necessary as we start to build complex patterns. While having more containers to separate layout from content is usually preferred it does add complexity to the markup. 

For example, what do we gain by having an `__action` container around the `button`? It allows for additional buttons or content down the line (maybe a help button next to close) but does that possibility warrant more markup and complexity. It's always about finding the right balance between flexibility and simplicity. It is often very difficult to keep things simple.

Let me consider how to best do this and we can circle back if need be.

LeoT7508:
@saadmhybris 

White text on orange doesn't pass color contrast accessibility, so we are going with dark text.

saadmhybris:
@LeoT7508 - I have changed the color to black, but still don't feel right to me as a person of poor eyesight. maybe we can test this at some point later. for now, it's ready for design review. 

@xak - your requested changes are in. please have a look.

LeoT7508:
@saadmhybris 

What branch can I view this in?

saadmhybris:
@LeoT7508 i just restored feature/885 for your review. The branch was deleted as it was reviewed and merged.

Issue #958, opened 2017-08-30
Pkg: Implement dropdown menu component


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `v2.0`

This is a new component for Techne that was previously supplied by Bootstrap

# Inputs
- See [1.5 implementation](https://techne.yaas.io/components/Contextual-Menu.html)

## Specs
- **Zeplin:** https://zpl.io/bzlRrMV
- **PSD**: https://app.box.com/folder/14968986610

## Basic block structure (suggested)

```
.tn-dropdown
```

# Outputs

- Match [Bootstrap](https://getbootstrap.com/docs/4.0/components/dropdowns/)

PR #965, opened 2017-09-05
Feature/958

xak:
#958


xak:
This is ready for review ... branch `feature/958`, go to http://localhost:8080/dropdown

The implementation matches https://zpl.io/bzlRrMV

LeoT7508:
@xak 

Drop downs are looking good. 

Side note:
I think the url you're using above is wrong. "http://localhost:8080/dropdown" wouldn't work. I changed 8080 to 3030 and it worked. FYI.

xak:
@LeoT7508 Yeah, it's always 3030. My mistake.

saadmhybris:
@xak  - a couple of questions. 

- the menu does not go away when you click elsewhere on the page.
- do we need to show any kind of validation with this? I know it's not a form element, but what if someone wants to use a form instead of select control?

xak:
@saadmhybris 

#1 I don't see much value in expanding the JS. This is minimal to test the functionality. We will need to discuss how much (if any) JS support we want to provide esp. since there will be no Bootstrap JS to do these small things. I'll open a separate JS story.

#2 Same with the validation. The button itself inherits all the form behaviors so it should get a green background with the `is-valid` class, etc. This dropdown has little value on it's own and is better suited for navigation dropdowns whereas Select2 or similar library would be better suited for form-like interactions. That needs to be addressed separately. I took pains to ensure the markup and structure match the Select2 outputs so it should be easy to expand this component to work with more advanced use cases.

saadmhybris:
@xak - ok makes sense. PR looks good from code perspective. I'll merge once @LeoT7508 give the green light.

Issue #969, opened 2017-09-05
Docs: Implement Alerts component


saadmhybris:
Alerts component being developed under #885. Once complete, implement documentation.

PR #972, opened 2017-09-06
#969: Docs - Impliment Alerts

saadmhybris:
#969


test/templates/icon/component.njk
```diff
@@ -4,5 +4,5 @@
     instead of more complex modifier.block options
 #}
 {%- macro icon(glyph="", size="", classes="") -%}
-<span class="tn-icon{{ (glyph if size == "" else [glyph, size]) | modifier('icon') }}{{ classes | classes }}" role="presentation"></span>
```
xak: This should stay. I added this for the dropdown based on what we had discussed. Removing it breaks the dropdown component.

saadmhybris: this is adding an empty class "tn-"

<span class="tn-icon tn-icon--add tn-" role="presentation"></span>

xak: This should be fixed with #978 — https://github.com/SAP/techne/pull/978/files#diff-ace3b50701c30da220db72a852e66f79R52

saadmhybris: ok will restore and merge.

Issue #964, opened 2017-09-05
Docs: tiny bug inside of the icon


like-me1:
There is tiny visible issue with the icon, see attached
![screen_shot_2017-09-05_at_14_47_39](https://user-images.githubusercontent.com/2720077/30061896-85323670-9249-11e7-8ec0-92d55ab08402.png)

PR #974, opened 2017-09-06
964: Bug - Fix Done Icon.

saadmhybris:
#964


saadmhybris:
@like-me1 - I resolved the conflict. can you please have a look.

Issue #916, opened 2017-08-25
Pkg: Implement contextual menu


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `v2.0`

This story is more about building a dropdown menu that can be used for contextual choices.

# Inputs
- See background at #907
- See [1.5 implementation](https://techne.yaas.io/components/Contextual-Menu.html)

## Specs
- **Zeplin:** https://zpl.io/VqQKNGV
- **PSD**: https://app.box.com/folder/14968986610

## Basic block structure (suggested)

```
.tn-dropdown
```

# Outputs

- Match 1.5 functionality

PR #975, opened 2017-09-06
#916 contextual menu

xak:
#916

### DIFF
--- a/scss/components/contextual-menu.scss
+++ b/scss/components/contextual-menu.scss
@@ -1,0 +1,16 @@
+@import "../core/settings";
+@import "../core/mixins";
+@import "../core/functions";
+
+/*!
+.tn-contextual-menu
+*/
+$block: ns(contextual-menu);
+
+.#{$block} {
+
+    margin-left: tn-space(5);
+
+
+
+}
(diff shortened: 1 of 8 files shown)


xak:
Ready for review ... branch `feature/916` at http://localhost:3030/contextual-menu

## NOTE: This is not a standalone component. You can use any type of button and a dropdown menu.

saadmhybris:
looks good. @LeoT7508 let me know when you are done reviewing and I'll merge this.

LeoT7508:
@saadmhybris 

This isn't correct. The icon is wrong. It should be vertical. And we don't have a "..." with "more" below it. Also when I click on the icon the menu opens but won't close. It just stays open. It should close when I click away from it.

xak:
@LeoT7508 I'm using the "more" icon we have now. http://sap.github.io/techne/icons.html. And the last example is the option you have for the Action Bar. You comped a "more" option there so I'm demoing that option.

LeoT7508:
@xak @saadmhybris 

Pending icon updates by me this is good to go.

saadmhybris:
@LeoT7508 we went back and forth on the more icon being verticle or horizontal in the past. we settled on horizontal because it allowed better visibility and hit area on mobile. You can check out the contextual menu page on 1.5 and it's horizontal - https://techne.yaas.io/components/Contextual-Menu.html

LeoT7508:
@saadmhybris 

We're going with vertical.

xak:
Is this OK?

LeoT7508:
@xak 

Yup. We should be good. I talked to Jeanne and Manu about this before when we decided to change it.

saadmhybris:
Sounds good on the vertical icon

Issue #888, opened 2017-08-21
Pkg: Implement action-bar component


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `dev`

# Inputs
- See background at #792
- See [1.5 implementation](https://techne.yaas.io/patterns/Action-Bar.html)

## Specs
- **PSD**: https://app.box.com/folder/14968986610
- **Zeplin:** https://zpl.io/bJJGLWb

## Basic block structure (suggested)

```
.tn-action-bar
	.tn-action-bar__navigation
	.tn-action-bar__title
	.tn-action-bar__actions
```

# Outputs

**There are some open issues with the Wizard UX, this intial implementation should only address the simple use cases.**

- Match 1.5 functionality

xak:
@LeoT7508 I need the back arrow icon to complete this ... the back icon in our current set is quite different.

![screen shot 2017-09-07 at 9 23 54 am](https://user-images.githubusercontent.com/424119/30168280-89bb2a56-93ae-11e7-8dce-97daa44be761.png)

PR #978, opened 2017-09-07
Feature/888

xak:
#888 

### NOTES
This implements basic uses cases for desktop browsers, shown are examples of a back button, page title, and page actions including a "More" contextual menu.

The icon font has also been recompiled to include the `backarrow` icon.


xak:
Ready for review in branch `feature/888` at http://localhost:3030/action-bar. See notes above.

saadmhybris:
screen shot 2017-09-07 at 8 24 40 pm

is the back arrow is still not visible?

xak:
@saadmhybris I'm seeing the new back arrow — https://github.com/SAP/techne/pull/978/files#diff-269648d1bd33699db8e308496fc0537dR22. Remove your local branch and check it out again. I had to force push some changes so maybe you have something old locally.

saadmhybris:
Looks good from code. 

on a side note, something wrong with the icons test page. it does not show the right icon. I'll open a ticket for it.

LeoT7508:
@xak 

Looks great. 

2 things:

1. What font weight are you using for the button labels?

2. I updated margins and padding for this to match the the Title Bar, Table, and Tree spacing. Should be a small update for this. 

Otherwise looks great.

xak:
@LeoT7508 The button text looks bold, and what margin and padding did you update?

LeoT7508:
@xak 

Ok, I think we can go with bold and see how it works. If it seems too heavy we can bring it down in weight. 

I updated the left & right margins, and the spacing from the back arrow icon and the page title. Its all up on Zeplin if you need to see it in more detail.

xak:
@LeoT7508 The problem with the 20px margin is that the large icon button (Back button) uses 28 pixels on each side. That's what the button has. So I either need to ...

1) pull the button over 8px to simulate the margin, 
2) create a special back button type just for this one button, or 
3) change all large icon button to use 20px padding left and right.

What do you propose?

LeoT7508:
@xak 

Well it depends on which section you want to patch together. Because if we don't use 20px padding for the action bar, then the icons in the title bar will have to be adjusted because that padding is 20px for those icons. Additionally, the icons and options in the tables and trees won't align if we use a different padding. So which creates less work for you is the real question?

xak:
@LeoT7508 The titlebar uses medium (standard) buttons which already have 20px. The tree I believe uses small buttons which use 16px.

LeoT7508:
@xak 

"The titlebar uses medium (standard) buttons which already have 20px. The tree I believe uses small buttons which have 16px."

This is correct.

xak:
@LeoT7508 I recommend changing the padding of large icon buttons to 20px. This is the only instance we have so far of this button. If I try to pull it over 8px the hit area will potentially overlap with the sidebar, which is not ideal.

screen shot 2017-09-08 at 11 51 50 am

saadmhybris:
this PR is ready to be merged once we get a green light from @LeoT7508

LeoT7508:
@xak 

We can change the icon large icon only buttons to 20px so 20px margins will work. 

All good here. Ill update my sheets and upload them to Zeplin.

xak:
@LeoT7508 The background color here should be it's own color in the palette. It's a major part of the page so it doesn't make sense to use a lighter version of another color. It's not intuitive that changing a neutral color would affect the background for a large part of page. 

The `lighter` and `darker` options should only be used when a base color is being modified, like for a `:hover` or `:active` state. That makes sense so when I create a new button with red text I don't have to then code separate values for the states. But in this instance the background _is_ the base color. This should be part of backgrounds.

LeoT7508:
@xak 

You are right about the color for the tool bar. I didn't use a percentage I added a background color I just didn't push it to Zeplin. Thats my miss, sorry about the confusion. The color is Background 2 #f0f5ff. 

About the percentages, what approach do you suggest to substitute for this?

xak:
@LeoT7508 For the `darken` and `lighten` functions we use now the differences are minor, 5% here, 10% there. I think if we stick to the hovers and active states it should be ok. I found a tool that shows the SASS functions — http://jackiebalzer.com/color — but its not very useful since you can't input a hex value.

I also tried several other SASS functions but none produced the exact colors in that hex color tool. I'll need to investigate more to see what method would work best or how you can simulate the SASS effects in Photoshop to accurately see the colors.

This was a surprised to me when I stated looking at it. Colors are complicated, apparently. 😄

LeoT7508:
@xak 

Ok so we will stick with percentages for the time being and try to find a solution. 

Thanks.

PR #981, opened 2017-09-08
Feauture/915 pagination component

like-me1:


test/templates/pagination/component.njk
```diff
@@ -0,0 +1,35 @@
+{#
+    see ./data.json for object structure
+#}
+
+{% macro pagination(properties={}, modifier={}, state={}, aria={}) -%}
+<ul class="tn-pagination">
+    <li class="tn-pagination--link-wrapper">
+      <a href="#" class="tn-pagination--link" aria-label="Previous">
```
xak: These should be `__` classes since they are elements, modifiers use `--`. 

Use `__item` instead of `__wrapper` since it's more semantically correct. When it makes sense, use names already widely used to describe relationship, see https://www.w3.org/TR/wai-aria/roles#listitem.

test/templates/pagination/component.njk
```diff
@@ -0,0 +1,35 @@
...
+    </li>
+    <li class="tn-pagination--link-wrapper">
+        <a href="#" aria-selected="true" class="tn-pagination--link">2</a>
+    </li>
+    <li class="tn-pagination--link-wrapper">
+        <a href="#" class="tn-pagination--link">3</a>
+    </li>
+    <li class="tn-pagination--link-wrapper">
+        <a href="#" class="tn-pagination--link tn-pagination--more" aria-label="Next">
+            <span aria-hidden="true">&#8226;&#8226;&#8226;</span>
```
xak: Add `role="presentation"` here.

test/templates/pagination/component.njk
```diff
@@ -0,0 +1,35 @@
...
+        <a href="#" class="tn-pagination--link">1</a>
+    </li>
+    <li class="tn-pagination--link-wrapper">
+        <a href="#" aria-selected="true" class="tn-pagination--link">2</a>
+    </li>
+    <li class="tn-pagination--link-wrapper">
+        <a href="#" class="tn-pagination--link">3</a>
+    </li>
+    <li class="tn-pagination--link-wrapper">
+        <a href="#" class="tn-pagination--link tn-pagination--more" aria-label="Next">
```
xak: You don't need a link here at all. This should be plain text inside the `__item`. The label is also incorrect.

test/templates/pagination/component.njk
```diff
@@ -0,0 +1,35 @@
+{#
+    see ./data.json for object structure
+#}
+
+{% macro pagination(properties={}, modifier={}, state={}, aria={}) -%}
```
xak: This is fine to rough out the elements, but think about building out the logic to render the component based on the data you supply. If you have 2345 items, with 30 per page and you are on page 28, what would you get. This is the best way to test the component — you can pass a variety of values to make you get the proper output.

saadmhybris:
Closing this since this PR is with the outdated branch. @like-me1 will open a new PR

Issue #985, opened 2017-09-11
Docs: Implement contextual menu


saadmhybris:
Component built under #916. Update drop-down components page to show the contextual menu

PR #990, opened 2017-09-11
#985: Docs - Implement contextual menu

saadmhybris:


joseegm:
@saadmhybris 
There is a small floating but with the other ActionBar buttons.
screen shot 2017-09-12 at 10 55 42

saadmhybris:
@joseegm - I removed the action bar documentation piece for now. the action bar is still under development so it's hard to make any adjustments to the component at the moment. It would just be that the markdown/liquid template is adding extra p tag, but lets' wait on the action bar documentation until that component is finished.

Issue #989, opened 2017-09-11
Docs: Update Actions Card example with contextual menu.


saadmhybris:
Contextual menu component is complete so we need to update the Actions Card example to show now it works.

PR #992, opened 2017-09-12
#989: update contextual menu on action card.

saadmhybris:
#989


joseegm:
Just removed the `href="#"` to avoid the page jumping around, please when used on the website avoid using it.

Issue #938, opened 2017-08-29
Bug: Bootstrap 3 causes minor alignment problems with spinner buttons


xak:
Seems to be related to font-size declarations. Each button should be vertically aligned.
 
![screen shot 2017-08-29 at 9 28 58 am](https://user-images.githubusercontent.com/424119/29826299-87a871ba-8c9c-11e7-9942-94fadbf9e3fd.png)

PR #993, opened 2017-09-12
#938: fix alignment problems with spinner buttons

saadmhybris:
#938 

Not sure if this is the best way to fix this, but this PR review will answer my questions I guess :)


saadmhybris:
@xak ready for review again.

Issue #955, opened 2017-08-30
Pkg: Implement Card Starter page


xak:
# Waiting for #954 

To better assist teams migrating to Techne 2.0, we will provide new starter pages. All components exist now for the Cards page. https://techne.yaas.io/starterpages/start.html

Some layout esp. responsive behaviors need to be filled in. Top and side nav can remain empty for now.

PR #1017, opened 2017-09-15
Feature/955

xak:
#955


xak:
@saadmhybris I made some modifications to the starter page ...

- Fix the icons (it was a simple routing fix in `app.js`
- Updated the cards to match the content at https://techne.yaas.io/starterpages/cards.html
- Updated card to use the `label` component
- Changed the `action-bar` background to match the toolbar (the correct one) and added borders (per the design)
- Added bottom margin to the `toolbar`
- Added the "description help" text and created some helpers to enable the layout (temp solution)

@LeoT7508 Please take a look in branch `feature/955` at localhost:3030/pages/cards. There are some things missing and not quite right that we should follow up on. Please open an issue with anything you see that needs to be fixed, like the "Add" button still needs to be done, the cards are padded too much and squish awkwardly, etc.

I think we're OK to release as long as we capture what is out of sync and circle back.

saadmhybris:
looks good, thanks!

LeoT7508:
@xak 

Looks like you deleted branch 955. Please let me know when its back up so I can check it out, thanks.

xak:
@LeoT7508 Sorry, @saadmhybris merged it so you can see it in branch `v2.0` now.

Issue #1029, opened 2017-09-18
Bug: Icons are broken all the components.


saadmhybris:
Icons are broken for all the pages. works only on the starter page. 

<img width="1375" alt="screen shot 2017-09-18 at 12 11 10 pm" src="https://user-images.githubusercontent.com/15215400/30554655-e1b32c44-9c6a-11e7-9fcb-738880caef37.png">

PR #1031, opened 2017-09-18
#1029 fixes icons font route

xak:
#1029


xak:
I went ahead and fixed a problem with the contextual menu ... the button CSS was missing.

screen shot 2017-09-18 at 2 53 39 pm

Issue #950, opened 2017-08-30
Pkg: Implement table component


xak:
# Workflow

- [x] Develop
- [ ] Design review
- [x] Code review
- [x] Merge to `v2.0`

# Inputs
- See background at #807 
- See [1.5 implementation](https://techne.yaas.io/patterns/Tables.html)

## Specs
- **Zeplin:** https://zpl.io/299YqA2
- **PSD**: https://app.box.com/folder/14968986610

## Basic block structure (suggested)

```
.tn-table
```

# Outputs

- Match 1.5 functionality

PR #1033, opened 2017-09-18
Feature/950

xak:
#950

### DIFF
--- a/scss/components/table.scss
+++ b/scss/components/table.scss
@@ -28,7 +28,6 @@ $block: ns(table);
         }
         &.is-selected,
         &[aria-selected="true"] {
-            background-color: tn-color(neutral, 3);
             background-color: lighten(tn-color(neutral, 3), 5);
         }
     }
--- a/scss/components/table.scss
+++ b/scss/components/table.scss
@@ -3,9 +3,10 @@
 @import "../core/functions";
 
 /*!
-.tn-table+(--no-border)
-    .tn-table__content+()
-    .tn-table__title+()
+.tn-table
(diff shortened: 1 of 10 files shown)


xak:
Ready for review in branch `feature/950` at http://localhost:3030/table.

NOTE: This is **ONLY** the views shown on [Zeplin](https://zpl.io/299YqA2), there is no responsive behavior.

The structure and CSS are not BEM compliant. This is intentional since table elements are unique and extra classes like `tn-table__row` and `tn-table__data` are redundant and can get tedious. This does increase the specificity but the trade-off in this case is worthwhile. The downside, however, is that these styles cannot be used outside of a table. If we have a use case where `div`s need to be styles as a table, we can circle back and add addt'l classes.

scss/components/table.scss
```diff
@@ -0,0 +1,60 @@
...
+    width: 100%;
+    max-width: 100%;
+    border-collapse: collapse;
+    tr {
+        &:hover {
+            background-color: lighten(tn-color(neutral, 3), 10);
+        }
+        &.is-selected,
+        &[aria-selected="true"] {
+            background-color: lighten(tn-color(neutral, 3), 5);
```
xak: @LeoT7508 The SASS lighten function did not return the exact colors you specified, so I matched as close as I could. I like using the same color - `neutral-3` - for hover and selected instead of different color, these shades come pretty close to what you have in Zeplin.

saadmhybris:
Restored branch to @LeoT7508  can review it. sorry got trigger happy again

Issue #1026, opened 2017-09-18
Docs: Implement Table starter page


saadmhybris:
component development under #1004

PR #1057, opened 2017-09-25
#1026: Docs - Implement table starter page.

saadmhybris:


xak:
@saadmhybris This looks good to me based on what we have now and I'll merge it. Moving forward I think @joseegm should be the reviewer for website content. That way he can create additional issues for content revisions and what not.

saadmhybris:
@xak, ok makes sense.

Issue #1067, opened 2017-09-27
Document modal


joseegm:
goal: Add more documentation and examples to the modal website component page.

PR #1068, opened 2017-09-27
#1067 Modal documentations and more examples added

joseegm:


saadmhybris:
Looks good, but at some point, we should have the tech writer's team look at this.

Issue #1005, opened 2017-09-14
Pkg: Add button for add collection item


xak:

saadmhybris:
wires/design under #1021

saadmhybris:
for MVP add a collections title and a + button as shown in wires.

PR #1069, opened 2017-09-28
#1005: new add button component + updated started pages with add button.

saadmhybris:


saadmhybris:
This button can we viewed at 

- http://localhost:3030/add-button
- http://localhost:3030/pages/cards
- http://localhost:3030/pages/table

xak:
Another note about the helpers ... for me, when we are forced to use a lot of helpers, it's a cue to think about that container and decide if we need to have it stand on it's own. For example, `.tn-container` could have a it's own header, actions, and footer with their own styles and behaviors. Actions always align right, header aligns left, footer is centered (e.g., "See All"). 

These are all the small details related to page-level UI and layout that we haven't spent much time on yet.

saadmhybris:
@xak i have made the changes as suggested.

Issue #1096, opened 2017-10-06
Create jekyll components index page


joseegm:
goal: create a dynamic index page using the info on the components sidebar yaml data

PR #1097, opened 2017-10-06
#1096 Added component cards to the components index page

joseegm:


saadmhybris:
Merging for now. Assuming we'll eventually have images for each of these component cards. maybe we can show a high-level skeleton of a component.

Issue #1030, opened 2017-09-18
Pkg: Icon updates


LeoT7508:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `dev`

All 1.5 icons have been updated to match the new design system. 
SVG files can be found on box here:
https://hybris.box.com/s/bzssyy0155zlvjezbct1mgbp944w0dnq

All new icons must pass through design before being added to this folder.

saadmhybris:
@LeoT7508  - can't seem to find the icon files at the link, can you pls double check? we need them to be SVG file of a consistent file which you already know...

PR #1099, opened 2017-10-06
#1030: Pkg - Icon updates

saadmhybris:
@LeoT7508 please take a look at this branch. 

there are a couple of icons that are not rendering correctly. They are named and shamed below. 

Found a polyline element in the icon "clone" the result could be different than expected.
Found a polyline element in the icon "edit" the result could be different than expected.

Also, I had to export some of the new icons as they were not compatible. I am attaching a screen of the SVG export settings for Illustrator that generally works.

svg export


saadmhybris:
@LeoT7508 - updated the cions that were causing issues. pls pull latest so see the changes 
@xak - this is ready for code review as well.

xak:
@saadmhybris Looks good if you can resolve the conflicts.

PR #1102, opened 2017-10-09
Feature/884 list group

like-me1:


scss/components/list-group.scss
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+  &__control{
+    flex-basis: $tn-list-group--input-width;
+    min-width: $tn-list-group--input-width;
+  }
+  &__label{
+    flex-basis: 95%;
+  }
+  &__button{
```
saadmhybris: We are creating a whole new button here instead of simply extending the existing button we have. Please refer to the[ alerts component](https://github.com/SAP/techne/blob/v2.0/scss/components/alert.scss#L31) where we have a similar action. Then we can [simply use the icon in the component structure](https://github.com/SAP/techne/blob/v2.0/test/templates/alert/component.njk#L10). this way, we are repurposing the existing components rather than creating a new one.

scss/components/list-group.scss
```diff
@@ -0,0 +1,81 @@
...
+$tn-list-group-hover: tn-color(action);
+$tn-list-group--input-width: tn-space(7) !default;
+
+.#{$block} {
+  @include reset;
+  vertical-align: top;
+  min-height: $tn-forms-height--input-text;
+
+  &__item{
+    border: 1px solid $tn-list-group--border-color;
```
saadmhybris: 1px is used four times. it's worth to put this in a var ($tn-list-group--border-width).

scss/components/list-group.scss
like-me1: @saadmhybris I simply re-used what we already have in place: http://sap.github.io/techne/input-group.html (Input Group — Special addon) why not exchanging it with new one everywhere when it's available since alerts example looks totally different from design prospect of view and why would we need different implementations for same button?

test/templates/list-group/data.json
```diff
@@ -0,0 +1,4 @@
+{
+    "type": "text",
+    "properties": []
```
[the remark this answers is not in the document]
like-me1: great hint learned!

test/templates/list-group/index.njk
```diff
@@ -0,0 +1,74 @@
...
+{{ list_group(
+    properties={
+          list: [
+          "List item 1",
+          "List item 2",
+          "List item 3 - Lorem ipsum dolor sit amet, consectetur sed do",
+          "List item 4 - Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud"
+          ],
+          modifier: {
+            after: icon("msgdanger")
```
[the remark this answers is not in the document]
like-me1: We can't ignore the checkboxes since the label next to checkbox, by default, aligned underneath if it's too long, and this is not what we need according to design. A tiny thing but changing the overall approach.

scss/components/list-group.scss
like-me1: @saadmhybris as per @xak suggestion we shouldn't care much about the content of the action, it could be delete/edit/sort etc. there even could be few of them. So I just created a container .tn-list-group__button which position any button int he right place. For the delete button I simply re-used .tn-alert__close, it's nothing for the list-group component itself. I also provided a form-related container tn-list-group__action-form which would be useful for checkboxes and radio buttons. This time I don't see the possibility simply re-use the form styling since the long label text needs some additional styling for the list-group, a bit different from what the form has (The label text can't go underneath of the checkbox or radio button, it needs to be positioned next to it with some padding)

Issue #951, opened 2017-08-30
Pkg: Implement tree component


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `v2.0`

# Inputs
- See background at #808 
- See [1.5 implementation](https://techne.yaas.io/patterns/Trees)
- See https://www.w3.org/TR/wai-aria/roles#tree

## Specs
- **Zeplin:** https://zpl.io/aXp6KXa
- **PSD**: https://app.box.com/folder/14968986610

## Basic block structure (suggested)

```
.tn-tree+(--header)
    .tn-tree__group+(--sublevel-1...-8, ([aria-hidden] | .is-hidden))
    .tn-tree__item
        .tn-tree__row+(--header, ([aria-selected] | .is-selected))
            .tn-tree__col+(--control, --actions)
                .tn-tree__control+([aria-pressed] | .is-pressed)
```

# Outputs

- Match 1.5 functionality

PR #1104, opened 2017-10-09
Feature/951

xak:
#951

### DIFF
--- a/scss/components/tree.scss
+++ b/scss/components/tree.scss
@@ -45,7 +45,7 @@ $block: ns(tree);
         border-top: solid 1px $tn-tree-border-color;
     }
     &__row {
-        padding: $tn-tree-cell-padding $tn-tree-cell-spacing;
+        padding: 0 $tn-tree-cell-spacing;
         display: flex;
         position: relative;
         &:hover {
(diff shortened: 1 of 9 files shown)


xak:
This is ready for review at branch `feature/951` at http://localhost:3030/tree

**Note:** I did not include an example of the selected state but it is available. Just add `aria-selected="true"` or class ```.is-selected` to the approriate `__row` element. The coding to implement this for demo-able purposes alone is substantial and since it is not present in the existing component I skipped it.

LeoT7508:
@xak 

The only issue I see here is the text doesn't seem to be centered vertically in the row. Theres less padding on above than below the text. 

screen shot 2017-10-09 at 2 31 47 pm

xak:
@LeoT7508 Fixed. Pull the changes and check again.

scss/components/tree.scss
```diff
@@ -0,0 +1,115 @@
...
+        //pad to account for control
+        padding-left: $_row_padding;
+        &--control {
+            flex: auto;
+            width: 25%;
+            //sets 8 sublevels of indentation based on __group level
+            @at-root {
+                @each $i in 1,2,3,4,5,6,7,8 {
+                    .#{$block}__group--sublevel-#{$i} & {
+                        //padding-left: ($tn-tree-cell-spacing * ($i + 1));
```
saadmhybris: we can get rid of this commented line if no longer needed

Issue #952, opened 2017-08-30
Pkg: Layout mixin


xak:
Should be able to include layout into components without using the layout classes.

A mixin allows layout to be attached to semantic classes. See https://github.com/SAP/techne/wiki/Layout-SCSS

PR #1107, opened 2017-10-10
Feature/952

xak:
#952

### DIFF
--- a/scss/components/form.scss
+++ b/scss/components/form.scss
@@ -29,7 +29,7 @@ $block: ns(form);
     //ELEMENTS *******************************************
     //set all ELEMENT baseline styles
     &__group {
-        @include clearfix;
+        @include tn-clearfix;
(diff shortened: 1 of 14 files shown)


xak:
This is ready. I put a page up at http://localhost:3030/layout with examples of both `.tn-container` and `tn-col`.

The mixin is part of `scss/layout/mixins`. The `scss/layout/col.scss` uses the mixins to generate the classes but it is not yet in use anywhere else.

See https://github.com/SAP/techne/wiki/Layout-SCSS for usage guidelines. Since we have no UX and design guidance for how the layout will be applied, we'll start with this implementation and add modifiers as we go forward. For example, the `.tn-container` defaults to a max-width with left and right padding built in, however, since table and tree component design have borders that bleed from edge to edge, maybe the default should be full width and max-width can be opted in. I don't know the best option.

Issue #1103, opened 2017-10-09
Fix Input with addons to expand to the full with


joseegm:
### Probem
The inputs with addons with doesn fit the 100% of the parent container

<img width="616" alt="screen shot 2017-10-09 at 15 49 00" src="https://user-images.githubusercontent.com/22662903/31341380-98e0d912-ad09-11e7-91e5-6ab146c0f783.png">

PR #1108, opened 2017-10-10
#1103 changes default to full width; adds inline option

xak:
#1103


xak:
@like-me1 The behavior of the `input-group` should match a plain `input` so we shouldn't ass side spacing. You would use it in a `form__item` which will give you a `label` and the correct side padding.

Issue #1079, opened 2017-09-29
Pkg: Implement Tree starter page


xak:
https://techne.yaas.io/starterpages/tree.html

PR #1110, opened 2017-10-10
#1079 tree starter page

xak:
#1079


xak:
See http://localhost:3030/pages/tree

Issue #1115, opened 2017-10-11
Pkg: Finalize app-level UI structure


xak:
Includes the app-level header, body, sidebar, footer — the primary containers for the site shell.

| block | elements |
|---|---|
| ui | __header, __app, __footer, __overlay |
| app | __sidebar, __main |

The UI layout should support ...
- A fully fixed UI with header, body and footer docked to the browser sides, sidebar and content areas should scroll independently
- A fixed header with scrolling content area
- Both above with no footer
- Both above with no sidebar
- Min heights such that footers do not float on short pages
- An `IFRAME` option for the main content area

xak:
![ui-fixed](https://user-images.githubusercontent.com/424119/31569745-df2d96ee-b042-11e7-8bc5-95768625cc54.png)

PR #1126, opened 2017-10-13
Feature/1115

xak:
#1115 

See https://github.com/SAP/techne/wiki/Page-Layout-Details

- Changes `tn-at-media` mixin to `tn-screen` for responsive media queries, updated all references in SCSS
- Adds `ui`, `app` and `overlay` layouts SCSS files
- The `ui` component deals only with top-level header, body and footer, as well as an overlay container to cover the entire screen
- The UI blocks can be fixed or static to create various template types, set either globally as SCSS vars or as classes on the elements. The `foo` theme sets a fixed UI for demo purposes.
- The `app` component defines the sidebar and main app content, there is no dependency on the UI and the `app` component could standalone.
- The app sidebar has some responsive behaviors built in (will need refinement and possible some JS assistance as we define what is inside the container†).
- The `overlay` is simple and should fill the entire container with centered content.

In `test/templates/pages/page.njk` set any or all of these to `true` to test variations of the layouts.

```
{% set hide_footer = false %}
{% set fix_header = false %}
{% set fix_footer = false %}

{% set show_overlay = false %}
```
 
Also, replace the commented line for the sidebar to show a full-width layout without a sidebar.

```
{% set hide_sidebar = true %}
```

These base containers intentionally have no padding built in. Child containers will need to do that since each of these must support the ability to bleed a container across the entire width, e.g, the top "back-to" or bottom "legal" section in the sidebar, the action bar and toolbars in the main content area.

_† For example, the app containers could be set to auto width and repsond to the content inside. The navigation inside the container could be minimized and expanded on hover whereas the `app` sidebar would simply pickup the width of the nav. Need to better understand the types of navigation available._

### DIFF
--- a/scss/all.scss
+++ b/scss/all.scss
@@ -2,6 +2,8 @@
 @import url("//fonts.googleapis.com/css?family=Dosis:400,500,600");
 @import url("//fonts.googleapis.com/css?family=Open+Sans:400,400italic,600,700");
 
+//@import "theme/foo";
+
 @import "core";
 @import "layout";
 @import "components";
--- a/scss/components/card-group.scss
+++ b/scss/components/card-group.scss
@@ -19,7 +19,7 @@ $block: ns(card-group);
     display: flex;
     flex-wrap: wrap;
     flex-direction: column;
-    @include tn-at-media(s) {
+    @include tn-screen(s) {
         flex-direction: row;
         & > * {
             flex: 1;
--- a/scss/components/toolbar.scss
+++ b/scss/components/toolbar.scss
@@ -53,7 +53,7 @@ $block: ns(toolbar);
             display: flex;
             align-items: center;
             border-top: solid 1px $tn-toolbar-border-color;
-            @include tn-at-media(s) {
+            @include tn-screen(s) {
                 border-top: none;
             }
         }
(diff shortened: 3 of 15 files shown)


xak:
This is ready for review in branch `feature/1115` at http://localhost:3030/pages/page

**NOTE**: These are most basic building blocks for the page layouts. The next step is to build out the `page` and `section` components that will hold the main content ... and the contents for the top and sidebar navs.

saadmhybris:
Here is a horizontal scroll on smaller screens size. this goes way on a bigger screen. 

screen shot 2017-10-15 at 12 51 08 pm

xak:
@saadmhybris I can't reproduce the horizontal scrollbar. What browser did you use? Is this the OOTB setup with fully scrolling content, i.e., **not** fixed header?

I can see there is extra space between the sidebar and main areas which should not be there. They should be flush. That's what is pushing the page to the right but I don't know where it's coming from.

screen shot 2017-10-16 at 8 47 29 am

LeoT7508:
@xak @saadmhybris 

I have been working with Jinjia to put together a spacing system for sections. I will create a guide for it today.

saadmhybris:
@xak I just double checked and looks like scrollbars are visible because of my mac settings to always show scroll bars. so seems like a non-issue unless someone else complains on say a windows machine or something.

Issue #917, opened 2017-08-25
Pkg: Implement tooltips component


xak:
# Workflow

- [ ] Develop
- [ ] Design review
- [ ] Code review
- [ ] Merge to `v2.0`

This is tooltip only. There is no JS enhancement available, this is on `:hover` only.

# Inputs
- See background at #812
- See [1.5 implementation](https://techne.yaas.io/components/Tooltip-and-Popover.html)

## Specs
- **Zeplin:** End of the page: https://zpl.io/2Gvzk9b
- **PSD**: https://app.box.com/folder/14968986610

## Basic block structure (suggested)

```
.tn-tooltip
```

# Outputs

- Match 1.5 functionality

saadmhybris:
blocked since we don't have designs for standalone tooltip component

saadmhybris:
how about 'Help Bubble' ? :)

PR #1128, opened 2017-10-15
#917: Pkg - Implement tooltips component

saadmhybris:
This feature can be found under branch - feature/917
This feature can be tested under - http://localhost:3030/tooltip


LeoT7508:
@saadmhybris 

This all looks good. 

Thanks.

test/templates/tooltip/component.njk
```diff
@@ -0,0 +1,15 @@
+<!--
+tooltip:
+    properties={},
+    modifier={ block: [] }
+-->
+{% macro tooltip(properties={}, modifier={}, state={}, aria={}) -%}
+<div class="tn-tooltip"{{ aria | aria }}>
+    {%- if modifier.block %}
+    <div class="{{ modifier.block | modifier('tooltip__arrow') }}"></div>
```
xak: This should have a `tn-tooltip__arrow` class as well. A modifier class cam only be used with the base class.

scss/components/tooltip.scss
```diff
@@ -0,0 +1,82 @@
...
+    &__arrow{
+        width: 0;
+        height: 0;
+        border-style: solid;
+        border-width: 0 $tn-tooltip-arrow-width $tn-tooltip-arrow-width $tn-tooltip-arrow-width;
+        border-color: transparent transparent $tn-tooltip-content-background-color transparent;
+        position: absolute;
+        display: none;
+        &--left{
+            @extend .tn-tooltip__arrow;
```
xak: These `@extend`s are not necessary since you should include the base class along with the modifier class. See `component.njk` below.

scss/components/tooltip.scss
```diff
@@ -0,0 +1,82 @@
...
+
+    //BLOCK BASE *******************************************
+    @include reset;
+    background: $tn-tooltip-content-background-color;
+    padding: $tn-tooltip-padding;
+    display: inline-block;
+    position: relative;
+
+    //ELEMENTS *******************************************
+    &__arrow{
```
xak: Did you try to do this with `:before` pseudo elements? May be more trouble than it's worth but it would reduce the markup considerably.

test/templates/tooltip/data.json
```diff
@@ -0,0 +1,17 @@
+{
+    "id": "tooltip",
+    "name": "Tooltip",
+    "properties": {
+        "content": ""
+    },
+    "modifier": {
+        "block": [ "left", "right", "top-left", "top-right", "botom-left", "bottom-right"]
```
xak: This should be `arrow` not `block` since the modification is occurring on the element and not the block. There are no block modifications AFAIK.

Also, `bottom-left` is spelled incorrectly.

saadmhybris:
@xak - this pr is ready again for the review. 

notes - 
* Intentionally left out top-left and top-right position as it is not possible to without JS as the height of the tooltip could vary. maybe this is possible with CSS var, but we can investigate later if needed 
* This component is also coupled with the help icon (?) we'll have to discuss how to extend this to make it a true tooltip that can be applied to any text string.

test/resources/controls.js
```diff
@@ -29,6 +29,17 @@
                     target.setAttribute("aria-hidden", isExpanded);
                     return;
                 }
+                //tooltip
+                var isTooltip = this.getAttribute("aria-hastooltip") === "true";
```
xak: I believe this is not used and can be removed. 

### FYI
`aria-hastooltip` is invalid. These attributes are a controlled set with specific meanings. They can't be made up. See https://www.w3.org/TR/wai-aria/states_and_properties#global_states

scss/components/tooltip.scss
```diff
@@ -0,0 +1,112 @@
...
+        }
+        &--bottom-left{
+            left: -$tn-tooltip-arrow-offset;
+            &::before{
+                top: -($tn-tooltip-arrow-offset);
+                left: $tn-tooltip-arrow-offset;
+
+            }
+        }
+        // &--bottom-right::before{
```
xak: This can go.

test/templates/form/index.njk
```diff
@@ -208,6 +208,25 @@
 
     })
 }}
+{{ form_group(
```
xak: I'm not seeing that this is doing anything. It looks (and behaves) like the example above it.

test/templates/tooltip/component.njk
```diff
@@ -0,0 +1,19 @@
+<!--
+tooltip:
+    properties={},
+    modifier={ block: [] }
+-->
+{% macro tooltip(properties={}, modifier={}, state={}, aria={}) -%}
+<span class="tn-tooltip"{{ aria | aria }}>
+    <span class="tn-tooltip__content{{ modifier.arrow | modifier('tooltip__content') }}">
```
xak: The modifier key should match the element exactly. It's confusing to use `arrow` yet it modifies `content`.

test/templates/tooltip/data.json
```diff
@@ -0,0 +1,10 @@
+{
+    "id": "tooltip",
+    "name": "Tooltip",
+    "properties": {
+        "content": ""
+    },
+    "modifier": {
+        "arrow": [ "left", "right", "botom-left", "bottom-right"]
```
xak: Should be `content` instead of `arrow` to match the element that is being modified.

xak:
@joseegm I recommend renaming this component to something more descriptive. Tooltip is a commonly understood concept and what we have here is specifically attached to a `?` indicator with a single use. The `tooltip` name is a misrepresentation and may be confusing.

Currently, these are referred to as `help-inline` with `help-inline--tooltip` and ` help-inline--section` options. I don't know the difference between those.

test/resources/controls.js
saadmhybris: 👍 

the right approach would be to use data attribute instead of aria, I suppose.

saadmhybris:
@xak - cleaned up the leftover code. sorry about that! 

We had to create this custom tooltip component to go with the help controls because the bootstrap tooltip was too complex and not very flexible to work with. hence we have to versions of the tooltips. one that comes with bootstrap and one that is custom built with the help controls.

joseegm:
@xak let's talk about renaming it on Monday, I cannot come up with a better name at this moment

saadmhybris:
change the name to 'inline help'

saadmhybris:
@xak @joseegm - the component has been renamed 'inline-help'

Issue #1137, opened 2017-10-17
Pkg: Fixes for Lists


joseegm:
### Problems
1. Right buttons with icons markup correction
2. Right icons should be blue

design: https://app.zeplin.io/project/598cce4ab8aced8f15a420a1/screen/599b0146642141e75c19e669

saadmhybris:
change the name to 'inline help'

PR #1154, opened 2017-10-19
#1137: Pkg - Fixes for Lists

saadmhybris:
@xak, I crossed reference alter and other components with similar action buttons on the right and the markup seems to be correct. I have to make some overwrites on the button and subsequent icon to match the design comps more closely.  Can you pls check if this is kosher.

this feels a bit dirty, let me know if there is a better way of doing this. 
https://github.com/SAP/techne/blob/feature/1137/scss/components/list-group.scss#L87


like-me1:
@saadmhybris I'd guess "edit" - icon is a bit missplaced to the left, instead of being centerd. Or is it only for me?
screen shot 2017-10-25 at 13 03 00

joseegm:
@saadmhybris 
Yes, there is a normal icon button like table row actions or tree row actions, those come automatically with the color, no need to change the background color to make the circle pattern.

joseegm:
@saadmhybris 
This is collateral damaged when the icon library got updated, the right icon (the solid circle with an X inside) was removed.

xak:
@saadmhybris Apologies for taking so long on this. 

We should **not** be making special modifications to existing components. The right side should be a separate (optional) container with a button inside. That should be a standard icon button, no height adjustment, no border radius, etc. If we want an "X" with a circle around it then that's what the icon should look like.

This entire component can be greatly simplified. See https://github.com/SAP/techne/tree/feature/884. We can avoid the addt'l containers/classes entirely, i.e., `tn-list-group__action` and `tn-list-group__action-form`. The list with the checkboxes can be a `tn-form__item--check` which already sets the checkbox and label relationship. There's no need to recreate that here for this special use case.

@joseegm There's not enough detail in #1137. Please include screenshots, links to Zeplin, etc. to better explain the bug.

joseegm:
@xak @saadmhybris 
The whole thing with the wrong markup is because as mentioned before the right icon was removed. 
Design is at https://app.zeplin.io/project/598cce4ab8aced8f15a420a1/screen/599b0146642141e75c19e669

saadmhybris:
@xak @joseegm - no problem. i'll refactor the code and let you know when it's ready for review.

saadmhybris:
@xak - ready for review again.

Issue #1006, opened 2017-09-14
Pkg: Finalize page-level UI structure


xak:
Includes the page-level intro content, how containers are stacked, how different layouts are implemented.

Consider content hierarchy ...

```
.tn-page.tn-app__main
    .tn-section
        .tn-container
```

| block | elements | |
|---|---|---|
| page | |
| section | __header, __actions, __navigation, __footer | Layout only, bottom margins, borders |
| container | __col | |
| panel | | |

PR #1155, opened 2017-10-20
Feature/1006

xak:
#1006


xak:
This adds new templates that can be used to construct various page types. Specific page-level implementations have been added to the index page ...

- http://localhost:3030/pages/page
- http://localhost:3030/pages/section

These are still rough but can be enhanced. Specifically the hover for the sidenav needs to be better and some padding will need to be adjusted ... but the structure is the main goal of this story. These containers will be used to build a variety of pages so they need to be flexible yet logical and easy to implement.

Issue #1112, opened 2017-10-11
Pkg: Fonts should be opt-in


xak:
Fonts are currently hard coded into `all.scss`. They should be explicitly set into a theme or fonts file so they can be replaced. Now they always get downloaded whether you want them or not.

Also, should set an explicit font stack for each font type.

joseegm:
This is good! We should include the fonts as part of the theme settings and not directly on the library. The library will work with the standard platform font.

xak:
@saadmhybris Since we are tying fonts to specific themes, I recommend creating a Techne theme file at `scss/theme/techne.scss`. In that file, only include the fonts and the `tn-fonts` variable declaration - remove the `!default` - from `_settings.scss`

In `_settings.scss`, change `tn-fonts` to the browser defaults only. Then in `all.scss` uncomment the theme line and set that as your theme file. This will essentially preserve the Techne look and feel which is intentionally buried in `core/_settings.scss` but separates the fonts.

I think this is better than hardcoding them in the starter HTML `head`.

PR #1161, opened 2017-10-23
#1112: Pkg - Fonts should be opt-in

saadmhybris:
#1112


scss/core/_settings.scss
```diff
@@ -40,10 +40,10 @@ $tn-type: (
 
 //fonts
 $tn-fonts: (
-    body: #{"'Open Sans'", sans-serif},
-    header: #{"Roboto", sans-serif},
-    code: #{"'Courier New'", monospace},
-) !default;
```
xak: The `!default` flag needs to stay here.

scss/theme/techne.scss
```diff
@@ -0,0 +1,8 @@
+@import url("//fonts.googleapis.com/css?family=Roboto:400,500,700");
+@import url("//fonts.googleapis.com/css?family=Open+Sans:400,400italic,600,700");
+
+$tn-fonts: (
+    body: #{"'Open Sans'", sans-serif},
+    header: #{"Roboto", sans-serif},
+    code: #{"'Courier New'", monospace},
+) !default;
```
xak: The `!default` flag should not be here.

saadmhybris:
@xak pls check again. default flag fixed.

Issue #1101, opened 2017-10-09
Docs: Implement inline-help component


saadmhybris:
component development under #917

PR #1162, opened 2017-10-23
#1101: Docs - Implement inline-help component

saadmhybris:
#1101 

the changes can be seen here - http://localhost:4000/form.html#inputs-inline-help-elements


saadmhybris:
@joseegm pls check again. sorry the helper class for alignment was missing.

Issue #1168, opened 2017-10-30
Pkg: Update starter pages with page layout structure


xak:
Updated layout is now available. Starter pages should be updated.

http://localhost:3030/pages/page
http://localhost:3030/pages/section
http://localhost:3030/layout

PR #1170, opened 2017-10-30
#1168 updates starter pages with new layout containers

xak:
#1168 

Updates all three starter pages (card, tree, and table) to use new layout containers. Also, includes a few small tweaks of various margins.

### DIFF
--- a/scss/components/card-group.scss
+++ b/scss/components/card-group.scss
@@ -19,7 +19,9 @@ $block: ns(card-group);
     display: flex;
     flex-wrap: wrap;
     flex-direction: column;
-    margin-bottom: -$tn-margin-bottom;
+    &:last-child {
+        margin-bottom: -$tn-margin-bottom;
+    }
     @include tn-screen(s) {
         flex-direction: row;
         & > * {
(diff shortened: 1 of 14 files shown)


xak:
@saadmhybris and @LeoT7508 Can you guys take a look at these? 

See branch `feature/1168`. Updated the three starter pages ...
http://localhost:3030/pages/table
http://localhost:3030/pages/tree
http://localhost:3030/pages/cards

These now use the ui, app, page, and section layout containers.

saadmhybris:
can be merged once @LeoT7508 is done reviewing.

LeoT7508:
@xak 

Table:
- Include the new 40px before 20px after a section. Looks like right now its 52px before. 
- The panel controls should have 16px below. 
- There should be a 1 px rule about the column headers. 

Tree:
- Include the new 40px before 20px after a section. Looks like right now its 52px before. 
- The panel controls should have 16px below. 

Cards:
- Include the new 40px before 20px after a section. Looks like right now its 52px before. 
- The panel controls should have 16px below. 

@saadmhybris 

We need to fix the icons, they were right for a while, now they aren't correct.

xak:
@LeoT7508 Fixed.

LeoT7508:
@xak @saadmhybris 

This checks out from the visual side. 

Thanks.

Issue #1174, opened 2017-10-31
Pkg: Fix tool bar icons.


saadmhybris:
- remove filter icon is missing 
- slight alignment issues on filer toggle icon. 
<img width="470" alt="screen shot 2017-10-31 at 11 15 29 am" src="https://user-images.githubusercontent.com/15215400/32235648-0bb88ba8-be2e-11e7-9364-3ac1410dcd35.png">

xak:
@saadmhybris The icons are fixed with PR #1170 —  https://github.com/SAP/techne/pull/1170/files#diff-b83e54035739031251bc3df22b6b2acb

The alignment is a related to the dropdown component. It looks worse without the Open Sans font.

saadmhybris:
@xak we can use this task to add the filter-remove icon back to match the original design. the close icon does not look good at all.

xak:
Sure.

Issue #1173, opened 2017-10-31
Docs: Form refactor


saadmhybris:
refer to comments in the PR - https://github.com/SAP/techne/pull/1169

PR #1183, opened 2017-11-01
#1174 adds back filterremove icon, updates docs icon page, fixes smal…

xak:
#1174 

This includes a new build of the dist package. Once merged, it will break `.tn-form` until #1173 is completed.


saadmhybris:
Looks good. thanks for updating the docs.

Issue #1011, opened 2017-09-14
Pkg: Implement tabs component


xak:
# Workflow

- [x] Develop
- [ ] Design review
- [ ] Code review
- [ ] Merge to `v2.0`

# Inputs
- See background at #1071 
- See [1.5 implementation](https://techne.yaas.io/components/Tabs.html)

## Specs
- **Zeplin:** https://zpl.io/VqygyvY

## Basic block structure (suggested)

```
.tn-tabs[role="tablist"]
    .tn-tabs__item
    .tn-tabs__link[role="tab"][aria-selected].is-selected[aria-disabled].is-disabled[aria-controls=#]
.tn-tabs__panel[role=tabpanel]#[aria-expanded].is-expanded
```
 > The panel element does not need to be inside the block container.

# Outputs

- Match 1.5 functionality
- Create horizontally scrollable container for small screens instead of a dropdown treatment
- Update JS to demo the tab functionality. See https://github.com/SAP/techne/blob/v2.0/test/resources/controls.js

PR #1195, opened 2017-11-03
#1011 implements tabs

xak:
#1011

### DIFF
--- a/scss/components.scss
+++ b/scss/components.scss
@@ -16,3 +16,4 @@
 @import "components/list-group";
 @import "components/inline-help";
 @import "components/nav";
+@import "components/tabs";
--- a/scss/components/tabs.scss
+++ b/scss/components/tabs.scss
@@ -4,7 +4,8 @@
 
 /*!
 .tn-tabs+()
-    .tn-tabs__item+([aria-selected]|.is-selected)
+    .tn-tabs__item?
+    .tn-tabs__link+((.is-selected|[aria-selected=true]),(.is-disabled|[aria-disabled=true]))
 .tn-tabs__panel+([aria-expanded]|.is-expanded)
 */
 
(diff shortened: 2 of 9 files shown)


xak:
Ready for review in `feature/1011` at http://localhost:3030/tabs

The two examples work the same just showing you can use different markup.

@saadmhybris Note: the JavaScript for this is separated from the rest. It's in the component `index.njk` file.

```
<script>
(function() {
    //climb up DOM to get block element
    function getBlock(control) {
        var block = control.parentNode;
        while (block.getAttribute("role") !== "tablist") {
            block = block.parentNode;
        }
        return block;
    }
    //get all tablists
    var tablists = document.querySelectorAll('[role="tablist"]');
    //handle each tablist
    for (var i = 0; i < tablists.length; i++) {
        var tablist = tablists[i];
        //get all the tabs
        var tabs = tablist.querySelectorAll("[aria-controls]");
        for (var j = 0; j < tabs.length; j++) {
            var el = tabs[j];
            el.addEventListener('click', function(e) {
                e.preventDefault();
                //check states
                var isDisabled = this.getAttribute("aria-disabled") === "true";
                var isSelected = this.getAttribute("aria-selected") === "true";
                if (isDisabled || isSelected) {
                    return;
                }
                //get all controls in grouping
                var block = getBlock(this);
                var controls = block.querySelectorAll("[aria-controls]");
                //handle each controls
                for (var k = 0; k < controls.length; k++) {
                    var control = controls[k];
                    //new selected tab
                    var isTriggerTab = control === this;
                    //get panel
                    var panelId = control.getAttribute("aria-controls");
                    var panel = document.getElementById(panelId);
                    //set states
                    control.setAttribute("aria-selected", isTriggerTab);
                    control.classList.remove('is-selected');
                    panel.setAttribute("aria-expanded", isTriggerTab);
                    panel.classList.remove('is-expanded');
                }
            })
        }
    }
})();
</script>
```

LeoT7508:
@xak 

It looks like the typeface is off here. I think you are using Helvetica. It should be Roboto. 

I looked at some of the other components to and it looks like its changed to Helvetica as well. Is this just a bug?

xak:
With #1112 the fonts are now part of the theme so the core component CSS falls back to the browser sans-serif default. There is no way around this but for the purpose of building the components the font doesn't matter much. The "Techne" theme will be shows on the website implementations.

Issue #1113, opened 2017-10-11
Pkg: Implement toggle buttons


xak:
# Workflow

- [x] Develop
- [ ] Design review
- [ ] Code review
- [ ] Merge to `v2.0`

# Inputs
- See [1.5 implementation](https://techne.yaas.io/components/Buttons.html)

## Specs
- **Zeplin:** https://zpl.io/25EQek0 zpl://screen?pid=598cce4ab8aced8f15a420a1&sid=59e617aaa633e7a9cf5817c2

## Basic block structure (suggested)

```
.tn-toggle
    input[type="checkbox""]
```

# Outputs

- Match 1.5 functionality
- Try to avoid styling the label, try the checkbox or the outer container.
- If this is unavoidable and the label is empty, be sure and include `aria-label`. All form fields must include some form of label for accessibility.

xak:
It is not possible to create a toggle without addt'l containers. I opted to use an empty container instead of attaching styles to the label. This makes it easier to reposition the label if desired and it is easier to completely swap a regular checkbox with a toggle  — no label required. The empty container is not ideal but its the best solution I could devise.

PR #1200, opened 2017-11-03
Feature/1113

xak:
#1113 

Ready for review in branch `feature/1113` at http://localhost:3030/toggle

**NOTE**: The color of the `checked` state is a little different from the design. I used the same color as the `checked` checkboxes instead of the darker one. Best to keep it the same for theming purposes. See https://github.com/SAP/techne/pull/1200/files#diff-e5af031203d514a0a6eab99b64f3d668R18

### DIFF
--- a/scss/components.scss
+++ b/scss/components.scss
@@ -17,3 +17,4 @@
 @import "components/inline-help";
 @import "components/nav";
 @import "components/tabs";
+@import "components/toggle";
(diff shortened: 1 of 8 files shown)


joseegm:
@xak 
One detail missing: Hover border effect like checkboxes and radio buttons is missing

xak:
@joseegm Hover state is there now.

screen shot 2017-11-06 at 10 42 07 am

Issue #1084, opened 2017-09-29
Pkg: Implement image/avatars


xak:
# <s>BLOCKED by #1175. Should use the default images (with absolute URLs) in all examples.</s>

# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `v2.0`

> NOTE: This a new component type.

The goal is easy-to-implement images with equal width and height with various border radius options regardless of the source image.

# Inputs
- See useful details from [1.5 implementation](https://techne.yaas.io/patterns/Images-Avatars.html) of images

## Specs
- **Zeplin:** 
https://zpl.io/VKGX07A
https://zpl.io/Vxm4nRE
https://zpl.io/VkDgmyv

## Basic block structure (suggested)
The squared corners option should be the default.
```
.tn-image((--rounded|--circle), (--xs|--s|--m|--l|--xl|--xxl), (--product|--profile))
```

# Outputs
- Must use a container with a background-image applied inline.

PR #1209, opened 2017-11-07
Feature/1084

xak:
#1084 — Ready for review in branch `feature/1084` at http://localhost:3030/image

## Design
- The abstract image is being pulled randomly from a (very slow) 3rd party site. Please provide the pixely image in the design if you want to use that one. The documentation can easily use a different image so it's not a blocker for this.  
- I increased the sizes of each to be unit of 4px. Not shown here, but I did test various images inline with various text sizes and the consistent 4px unit kept everything lined up more nicely.

## Dev
- The type sizes are updated here because there were some problems in Safari rounding the line-heights down. The `1` size, for example, was rounding down to 23px. This is mainly due to the hierarchy we have which results in fractional units. I tweaked the line-height values to render more consistent results.
- The placeholder SVGs are in `test/resources`. To generate the base64 data uri, run `svgo --datauri=base64 source_file.svg -o datauri_file.txt` then update the data URI var in `image.scss`. NOTE: SVGO must be installed globally - `npm install -g svgo`

## Documentation
- It should be noted on the site that the source images should be appropriately sized — usually no more than 2-4x larger than the display size. If using a `xs` icon, the image should be no more than 80px. This is not hard and fast, but we want to avoid shrinking large images down to small sizes for performance reasons.


LeoT7508:
@xak @saadmhybris 

Looks pretty straight forward to me. 

All good.

Issue #1076, opened 2017-09-29
Pkg: Implement Details 2 starter page


xak:
# Workflow

- [x] Develop
- [x] Design review
- [x] Code review
- [x] Merge to `v2.0`

https://techne.yaas.io/starterpages/details-2.html

https://zpl.io/25EXgKW

PR #1210, opened 2017-11-07
#1076 adds details 2 starter page

xak:
#1076 Ready for review at branch `feature/1076` at http://localhost:3030/pages/details-2

**NOTES**
- The tabs do not work, there is no content in the other tab panels even though they are in the markup.
- Reduced the height of the textarea to `52px*2` from `*3`. This was never specified in the form field designs.
- Added a bottom margin to `tn-col`s when they stack at small screens, they do not have this at larger sizes.
-  Added a bottom border to sections by default. The last item does not get one and there is a modifier — `tn-section--no-border` when it needs to be removed by force.

### DIFF
--- a/scss/components/form.scss
+++ b/scss/components/form.scss
@@ -22,7 +22,7 @@ $block: ns(form);
     $tn-form-message-font-size: -1 !default;
     $tn-form-message-color: tn-color(text, 2) !default;
     $tn-form-item-gutter: $tn-width--gutter !default;
-    $tn-form-item-margin-botton: $tn-margin-bottom !default;
+    $tn-form-item-margin-bottom: tn-space(8) !default;
 
     &__group {
         @include tn-clearfix;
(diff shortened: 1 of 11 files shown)


LeoT7508:
@xak 

It looks like we have some inconsistencies happening here. I think Jinjia confused some details. The guide for the sections can be found here:
https://zpl.io/VQ0lywx

Let me look through some files and see if I can make this more clear.

xak:
@LeoT7508 I'm pretty sure this matches the guides you reference. 40 on top, 20 on the bottom and sides.

screen shot 2017-11-09 at 11 08 09 am

What's missing?

xak:
See #1220

LeoT7508:
@xak 

The 2 main changes here are making the divider rule 2px tall & making the space between fields 32px, right now its 20px.

xak:
@LeoT7508 Both have been adjusted.

LeoT7508:
@xak 

Looks good to go from my end. 

Thanks for the updates.

PR #1211, opened 2017-11-07
Feature/1081

xak:
#1081 — Ready for review in branch `feature/1081` at http://localhost:3030/spinner

**NOTE**
- We should revisit all block containers — any that can contain content — that can be "loading". Components like `tn-section`, `tn-panel`, `tn-card-group`, `tn-card`, `tn-table`, `tn-tree` need a minimum height for their "loading" state. Almost every component should have some exploration around the [nine states ](https://medium.com/swlh/the-nine-states-of-design-5bfe9b3d6d85). Right now, a section that is loading would have no height.
- I did not create an example of the loading full page but it would work exactly like any other component by applying the busy attribute (or `is-busy` class) and showing the spinner.

```
<article class="tn-page" aria-busy="true">
    <div class="tn-spinner" aria-hidden="false" aria-label="Loading">
        <div></div>
    </div>
    ...
</article>
```
See it works on the Cards page ...

screen shot 2017-11-07 at 3 43 16 pm

I think that's it.


saadmhybris:
👍 for 9 states for design. 

@LeoT7508 @joseegm  - did we want to have different sizes and colors for this? I did not see a zeplin design so we only have one version so far.

joseegm:
@saadmhybris 
Nope. For now, to simplify things we only have one. This one, it will be used on the components that Zack mentioned.

LeoT7508:
@xak 

Is there a way we can show how it will looks when something is clicked? What will the transition be from click to finish load? What you are showing works, but it would be good to see the process from start to finish. Is this do-able?

saadmhybris:
@LeoT7508 - good thought we can mock that in a starter page perhaps to make them more interactive. i am merging this for now as the component looks good.

xak:
@LeoT7508 I'm not sure what you want to see. What do you mean when "something is clicked"?

I think the most common use case is an empty container that is being loaded. It could be an entire table, a single card, a group of panels, etc. Like this ...

screen shot 2017-11-06 at 1 21 27 pm

The way it would work now is that a panel, for example, would be the height of the spinner, when the content loads in the panel will jump to the height of the content. There's no transition, the spinner will be there, then it will be gone.

LeoT7508:
@xak 

What I'm saying is it would be good to see how full interaction works. 

Example:

Theres a normal page, you click on an action and then you see the loading spinner appear, spin, finish spinning, then disappear. The whole process from click, load, finish load, fade away.

xak:
See #1220

PR #1227, opened 2017-11-09
#1077: added empty starter page + couple more helpers.

saadmhybris:


test/templates/pages/empty.njk
```diff
@@ -0,0 +1,26 @@
+{% extends "./layouts/page.njk" %}
```
xak: There is a special template set up for starter pages that remove some of the app-level containers not required for these "module" pages. See https://github.com/SAP/techne/blob/v2.0/test/templates/pages/tree.njk

```
{% extends "./layouts/module.njk" %}
```

The template includes some variables you can set to output the common page elements. The `is_landing_page` determines whether the back button is shown or not.

```
{% set page_title = "Tree" %}
{% set page_intro = "Page level help message goes here" %}
{% set is_landing_page = false %}
{% set show_toolbar = false %}
```

All you have to put in is the `page_content`. The rest is done for you.

test/templates/pages/empty.njk
```diff
@@ -0,0 +1,26 @@
...
+<header class="tn-action-bar">
+    <h1 class="tn-action-bar__title">
+        Page Title
+    </h1>
+</header>
+{%- endblock %}
+
+{% block page_content %}
+
+{%- call layout.section(modifier="full-bleed") -%}
```
xak: You don't need `full-bleed` here.

test/templates/pages/empty.njk
```diff
@@ -0,0 +1,26 @@
...
+        <p class="tn-has-text-align-center tn-has-text-italic">
+            <span class="tn-has-text-bold">There are no items on this page.</span><br>
```
xak: Just use `strong` here instead. It's a legitimate use of the element. The italic helper is OK in this instance since the usage is more decorative than semantic.

scss/helpers/_general.scss
```diff
@@ -6,6 +6,12 @@ $ns: ns();
 .#{$ns}text-transform-none {
     text-transform: none !important;
 }
+.#{$ns}has-text-bold {
```
xak: These should be `has-font-weight-bold` and `has-font-style-italic`. With a few exceptions, the helpers use the exact CSS declaration terminology.

xak:
@saadmhybris I requested a review of the design from @LeoT7508 — https://zpl.io/bAmyq0P

It's not a blocker but we can do better than the "click here" link. An empty page is a deadend so the action should be prominent. There are only two things you can do here, go back or add something new. That tiny, tiny link is bad for the user. I recommend putting an "+ Add New Item" button below the text.

saadmhybris:
@xak all requested changes are now pushed. added "+ Add" button as well as suggested. @LeoT7508 is cool with this, we can merge this.

xak:
OK with me.

LeoT7508:
@xak 

I kept this really simple, its using the standard spacing, button styles and sizes. If we want to make something different we will have to create special cases for all this. Let me know if thats something we'd like to explore. 

stater page - table - 3 column 1580 no content

saadmhybris:
screen shot 2017-11-14 at 1 02 54 pm


I have updated the add button. note that i used the only add icon we have. we can update it later once we have the new add icon with round fill background.

xak:
Good with me!

Issue #1276, opened 2017-12-04
Pkg: Distribute theme directory


xak:
When the package is built the `theme` directory is not included.

Trying to use the `all.scss` as part of a custom SASS build returns an error since it's missing.

```
[INFO] Compiled 1 files
[ERROR] Error: File to import not found or unreadable: theme/techne.
       Parent style sheet: /Users/i860395/Workspace/automation-engine/cloud-portal/node_modules/techne/scss/all.scss
        on line 1 of node_modules/techne/scss/all.scss
>> @import "theme/techne";
```

This is occurring since we moved the fonts to `theme`

PR #1292, opened 2017-12-12
#1276 adds theme dir to npm package

xak:
#1276


saadmhybris:
tested with `npm pack` and it works.

Issue #1271, opened 2017-12-04
Site: Implement Homepage Design


saadmhybris:
Design - https://zpl.io/aXK0oPx
related ticket - #1194

PR #1295, opened 2017-12-14
Feature/1271

saadmhybris:
#1271 

may need to update the update the support links (have a design request? and ask a question.) base on feedback from the design comp on zeplin.


joseegm:
Let's try to center the main content later on. It is fine for now.

Issue #1308, opened 2018-01-03
Type page update


gcho1023:
Remove "a" in the sentence..."They are a timeless fonts..."
<img width="1062" alt="screen shot 2018-01-03 at 12 02 42 pm" src="https://user-images.githubusercontent.com/32933889/34533069-21b0027e-f07e-11e7-92ca-3fa5a9230b9a.png">

PR #1309, opened 2018-01-05
#1308 : copy update on type page

saadmhybris:
#1308


saadmhybris:
self-merging since this is a low-risk copy update.

